PR #6, opened 2014-03-25
Added another exception type to handle poorly formatted JWT input

mapitman:
Also let ReSharper clean up a few things


mapitman:
Someone else submitted a similar change

PR #7, opened 2014-04-29
Deserialze to object

jchannon:
I've added a method to deserialize to an object which you will then be able to have something like:

```
public class JwtToken
{
    public string Issuer { get; set; }
    public string Audience { get; set; }
    public IEnumerable<Claim> Claims { get; set; }
    public DateTime Expiry { get; set; }
}
```


mikelehen:
Cool!  Merged.

PR #11, opened 2015-02-12
Portable version of JWT

devinrader:
Created a new Portable project for JWT.  Had to pull in some external project dependencies to make up for missing native portable libraries :(

No tests in this or the non-portable project so I assume this works.


mikelehen:
Eh...  this looks like it creates a new copy of all of the code?  That's gonna' be a maintenance burden that I don't want. :-)

If you want this merged, can you explain more what the goal is here and your approach?  Are you trying to use this on Windows Phone or something?

I don't know much about Portable Class Libraries, but maybe we should just convert the whole project to be a Portable Class Library and it'll work everywhere?  Or is there some reason we'd actually want two separate projects?

devinrader:
Hmmm..not sure why git thinks the entire file changed.  Maybe line endings?  I can go back and see about fixing that.

The reason for the portable library is mostly to support the Xamarin platform, but also to reduce the maintenance costs in a new Twilio helper library that needs to support x-plat.

Its your call if you want to convert the whole library or not.  I'm happy to maintain this branch and build a seperate portable nuget package.  

The biggest drawback I see with supporting portable is that it requires external dependencies since there is no portable version of System.Security or of the built-in JSON decoder.

PR #17, opened 2015-05-03
Replacing JavaScriptSerializer with Json.Net

abatishchev:
In my opinion, it's better to use near-standard Json.Net than near-obsolete JavaScriptSerializer.
Also it will make it more portable - Json.Net is already is, plus no dependencies on System.Web.*.
What do you think?


abatishchev:
It may be a breaking change though. Bump minor or major version then? Maybe not now but as part of adding the RS256 support.

devinrader:
Implemented by #25

PR #18, opened 2015-05-04
Support Asymmetric Signing (especially RS256)

abatishchev:
Adding the support of RS256 using X509Certificate2.


justinhelgerson:
I realize this is a dumb question on my part, but I'm not very familiar with the Crypto namespace.  How do you create a JWT using your asymmetric additions?

abatishchev:
I don't remember exactly but from what I remember my fundings started there:
- http://stackoverflow.com/a/10106800/41956
- https://zavitax.wordpress.com/2012/12/17/logging-in-with-google-service-account-in-c-jwt/
- https://github.com/auth0/webapi-jsonwebtoken
- https://github.com/googleplus/gplus-verifytoken-csharp

abatishchev:
I don't remember exactly but from what I remember my fundings started there:
- http://stackoverflow.com/a/10106800/41956
- https://zavitax.wordpress.com/2012/12/17/logging-in-with-google-service-account-in-c-jwt/
- https://github.com/auth0/webapi-jsonwebtoken
- https://github.com/googleplus/gplus-verifytoken-csharp

abatishchev:
So here the question is how do you generate the signature (the 3rd segment) from the 1st (header) and 2nd (claimset). Because that's how Google OAuth2 powered services will validate the token provided.

abatishchev:
And from the implementation standpoint, the signature is eventually generated by this line: https://github.com/abatishchev/jwt/blob/rs256-1/JWT/JWT.cs#L47

justinhelgerson:
Sorry, I don't think I was clear with my question.  I was more interested in an example of how you would make a JWT with your asymmetric changes (an example like the home page of this project).  In particular the following line was confusing:  https://github.com/abatishchev/jwt/blob/rs256-1/JWT/JWT.cs#L39.  But, now that I look at it closer I presume the `byte[] key` comes from previously exporting a certificate (e.g. https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2.export%28v=vs.110%29.aspx)?

abatishchev:
Any chance to review this please?

JWT/JWT.cs
```diff
@@ -18,17 +20,31 @@ public enum JwtHashAlgorithm
...
 
         static JsonWebToken()
         {
-            HashAlgorithms = new Dictionary<JwtHashAlgorithm, Func<byte[], byte[], byte[]>>
+            HashAlgorithms = new Dictionary<JwtHashAlgorithm, Func<byte[], string, byte[], byte[]>>
+            {
+                { JwtHashAlgorithm.HS256, (key, _, value) => { using (var sha = new HMACSHA256(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.HS384, (key, _, value) => { using (var sha = new HMACSHA384(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.HS512, (key, _, value) => { using (var sha = new HMACSHA512(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.RS256, (key, secret, value) => RS256(key, secret, value) }
```
mikelehen: Maybe we should separate this into symmetric and asymmetric algorithms, or leave this as-is and treat RS256 as a special case since it needs different arguments.

JWT/JWT.cs
```diff
@@ -18,17 +20,31 @@ public enum JwtHashAlgorithm
...
+        {
+            var cert = new X509Certificate2(key, secret);
+            var rsa = (RSACryptoServiceProvider)cert.PrivateKey;
+            var param = new CspParameters
             {
-                { JwtHashAlgorithm.HS256, (key, value) => { using (var sha = new HMACSHA256(key)) { return sha.ComputeHash(value); } } },
-                { JwtHashAlgorithm.HS384, (key, value) => { using (var sha = new HMACSHA384(key)) { return sha.ComputeHash(value); } } },
-                { JwtHashAlgorithm.HS512, (key, value) => { using (var sha = new HMACSHA512(key)) { return sha.ComputeHash(value); } } }
+                KeyContainerName = rsa.CspKeyContainerInfo.KeyContainerName,
+                KeyNumber = rsa.CspKeyContainerInfo.KeyNumber == KeyNumber.Exchange ? 1 : 2
```
mikelehen: I don't have much experience with the .Net crypto libraries, but surely we have to do more than just get the KeyContainerName and the KeyNumber from the private cert for us to successfully encode something with it...?  But maybe I'm missing something.

JWT/JWT.cs
```diff
@@ -18,17 +20,31 @@ public enum JwtHashAlgorithm
...
+                { JwtHashAlgorithm.HS256, (key, _, value) => { using (var sha = new HMACSHA256(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.HS384, (key, _, value) => { using (var sha = new HMACSHA384(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.HS512, (key, _, value) => { using (var sha = new HMACSHA512(key)) { return sha.ComputeHash(value); } } },
+                { JwtHashAlgorithm.RS256, (key, secret, value) => RS256(key, secret, value) }
+            };
+        }
+
+        private static byte[] RS256(byte[] key, string secret, byte[] value)
+        {
+            var cert = new X509Certificate2(key, secret);
```
mikelehen: So the "key" here is actually a byte array containing a certificate, I guess?  What format does it need to be in?  Is it a custom serialization specific to the X509Certificate class or is it some standard certificate container format or something?

Perhaps we should let folks pass in a X509Certificate2 (or just the private key?) rather than a key and a secret.  That way they can get the cert however they want (and they only have to pass in one thing instead of two).

mikelehen:
Sorry I'm so slow!  I took a look, but I'm not sure about a few things...  I added some comments, but also:

* We should have an example in README.md of how you use this functionality, else nobody will know it's there or how to use it.
* It would be nice if we supported decoding asymmetric JWT's in addition to encoding them.
* Because I was so slow, there are merge conflicts.  Sorry!
* Somebody added some tests to the repo since you made this change.  Can we add a test for this?

JWT/JWT.cs
abatishchev: Frankly I don't have much either. I mentioned above what investigation did first and then found this code and made it working with Google Service Account. It definitely does :) With Google OAuth2 directly and via Auth0 (have your heard about it before?).

JWT/JWT.cs
abatishchev: My goal was to introduce as less changes as possible to increase the chances them to be reviewed and merged.
I think we could leave it as is and separate later once other RS* are added.

JWT/JWT.cs
abatishchev: Private key is a byte array in base64 representation (usually kept in Web.config for server-side web apps, you know). See [Google](https://developers.google.com/identity/protocols/OAuth2ServiceAccount) or [Auth0](https://auth0.com/docs/apiv2) for examples.

Secret (or password in term of MSDN) is just a string, see [X509Certificate2 ctor](https://msdn.microsoft.com/en-us/library/ms148418%28v=vs.110%29.aspx). For instance, for Google Service Account it's "notasecret", for client-side apps or at Auth0 it's something meaningful.

I think we can have an overload accepting either X509Certificate2, or byte[] and string and construct a cert by ourselves.
But such method will be usable only with RS256 and not with other algorithms.
On other hand, existing overload accepting secret is only usable with RS425 as well.

abatishchev:
- I'll add an example once we're done with the code, ok? Just in case we'll introduce changes to it. In addition, I'm not really good in writing docs in English, I'm not a native speaker.
- I'll take a look how to implement the decoding using RS256.
- I wish I knew how to merge your master into my branch. Then I could resolve conflicts, make sure all tests are passing and add more of them.

abatishchev:
Please review and merge #18, then I'll fix the reset of the conflicts.

abatishchev:
I merged from the latest upstream/master, updated the overloads of Encode/Decode, also their xml docs.

devinrader:
@abatishchev think there is any way to avoid having to pass the null as a method param in the overloads?  Feels a bit janky to me.  Other than that I'm cool with merging this in.

abatishchev:
Any call to `Encode()` ends up with calling `HashAlgorithms[algorithm](key, secret, bytesToSign)` so you need to pass either `secret` or `null`.

Or if you mean from an end-user perspective calling one of the overloads of `JsonWebToken.Encode()` then I could miss one of them, let me check, I'll update the branch.

abatishchev:
Oh, I'm tired merging conflicts. They're caused my overloads reordering. I merged from upstream/master several times. I'm not an expert in git merge though. I would just overwrite existing  Jwt.cs, i.e. take source when merging into master).

Other option would be extract Encode() into Jwt.Encode.cs and Decode() into Jwt.Decode.cs to simplify possible merging.

abatishchev:
I'm not sure in the last change thought. I tried to make it look like pluggable JSON serializer but still

What actually I would do is in ver. 2.0 move from the static class paradigm to an instance with dependencies. So you could inject necessary encoder/decoder, serializer or certificate factory.

Thoughts?

abatishchev:
For simplified transition, existing static class would look like this:

```csharp
public class JsonWebToken
{
  private static JsonWebToken _default = CreateDefault();

  public JsonWebToken(...)
  {
  }

  public static string Encode(...)
  {
    return _default.Encode(...);
  }

  public static string Decode(...)
  {
    return _default.Decode(...);
  }
}
```

abatishchev:
I'm not sure in the design of the last commit thought. I tried to make it similar to pluggable json serialzier but still.

abatishchev:
Please take a look.

abatishchev:
Any chance this to be reviewed, please?

abatishchev:
Any chance this to be reviewed, please?

abatishchev:
Any chance this to be reviewed, please?

sltkr:
Just wanted to add my support, I am super interested in an Asymmetric key implementation (RS256) like the one abatishchev is/was working on. Would it be possible to get this approved?

gouravd:
This is badly needed, now that Firebase goes all Google and to use Digits, a proper JWT is required

abatishchev:
Any chance to come back on this?

abatishchev:
Everybody who's interested in this change are welcome to take a look to the latest draft

msallin:
Hi

As I used parts of your code for my current project, I want to contribute back the new stuff.
I implemented the verification for a JWT signed with the RS256 algorithm.

The content is the Header + "." + Payload.

```
private static void VerifySignature(string content, byte[] crypto)
{
    ** // This dependency must be extracted for your lib. **
    ** X509Certificate2 x509Certificate2 = new X509Certificate2(certPath); **
    byte[] bytesToSign = Encoding.UTF8.GetBytes(content);
    SHA256 sha256 = SHA256.Create();
    byte[] hash = sha256.ComputeHash(bytesToSign);

    RSAPKCS1SignatureDeformatter rsaDeformatter = new RSAPKCS1SignatureDeformatter(x509Certificate2.PublicKey.Key);
    rsaDeformatter.SetHashAlgorithm("SHA256");
    if (rsaDeformatter.VerifySignature(hash, crypto))
    {
        return;
    }

    throw new SignatureVerificationException("Invalid signature.");
}
```

My suggestion to you is, that you abstract away the dependency to the used algorithm by overloading a class to the Encode/Decode method which represents the algorithm and provides a common interface to do the sign/verify stuff (similar like its done in the implementation from MSFT (see https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt/).

abatishchev:
If shortly, this is awesome, tanks for contributing back!

There are few issues though:
* This method doesn't conform the existing signature of `Verify()` at https://github.com/jwt-dotnet/jwt/blob/master/src/JWT/JWT.cs#L271
* So indeed requires an abstraction. What is always planned to version 2.0 but neither me nor anybody else didn't find time to implement all this staff, see https://github.com/jwt-dotnet/jwt/issues/55

abatishchev:
Closed in favor of https://github.com/jwt-dotnet/jwt/pull/68

PR #21, opened 2015-05-12
Pluggable JSON Serialization and Logging

andyday:
I currently use ServicesStack.Text for JSON Serialization and I know others who prefer Newtonsoft.Json.  I've added the ability to customize the provider used.

During development and debugging, I find it helpful to log the token JSON during encode and decode.  I've also added the ability to configure a Log method if logging is desired.

### DIFF
--- a/JWT.Tests/DecodeTests.cs
+++ b/JWT.Tests/DecodeTests.cs
@@ -1,0 +1,96 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Web.Script.Serialization;
+using FluentAssertions;
+
+namespace JWT.Tests
+{
+    [TestClass]
+    
+    public class DecodeTests
+    {
+        JavaScriptSerializer jsonSerializer = new JavaScriptSerializer();
+        string token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJGaXJzdE5hbWUiOiJCb2IiLCJBZ2UiOjM3fQ.cr0xw8c_HKzhFBMQrseSPGoJ0NPlRp_3BKzP96jwBdY";
+        Customer customer = new Customer() { FirstName = "Bob", Age = 37 };
+
+        [TestMethod]
+        public void Should_Decode_Token_To_Json_Encoded_String()
+        {
+            var expected_payload = jsonSerializer.Serialize(customer);
+
+            string decoded_payload = JWT.JsonWebToken.Decode(token, "ABC", false);
+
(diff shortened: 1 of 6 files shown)


abatishchev:
You may want to squash/combine commits into 1, or something like this.

devinrader:
I actually prefer this design over switching outright to and taking the dependency on JSON.NET which brings it own set of problems with it.  If you look at the history of the library JSON.NET was actually deliberately removed as a dependency a while ago.

@andyday we actually have tests now :astonished: so it would be awesome if you could pull those down and make sure they continue to pass and add where needed for this PR.

andyday:
@abatishchev you can see the summary of the change under the Files Changed tab.  there should be now need to consolidate commits.

abatishchev:
The point is to do not have multiple commits in the repo history, especially if later fixes problems cause but former.

andyday:
@abatishchev point taken.  I'll consolidate and add unit tests...

abatishchev:
Maybe rather than adding an interface, use `Func<string, object>` and `Func<object, string>`?

The very important reason to do not use System.Web.dll's JavaScriptSerializer is its problem with portability and future support of .NET Core. Completely unnecessary and importable dependency.

andyday:
I had thought about that.  Have a Serialize and Deserialize static property...  

Yes I'm not a fan of JavaScriptSerializer nor JSON.NET for that matter.  ServiceStack.Text has been a pleasure to use...  One particular point as it relates to JWT is its default handling of null properties by removing them from the serialized JSON.  This is helpful in reducing the token size.

abatishchev:
I never had issues with Json.Net and from my experience it has been a pleasure to use as well.

Plus ASP.NET Web API depends on it so using another Json serializer doesn't makes sense in this case.

andyday:
I don't use ASP.NET Web API.  I use ServiceStack.  My biggest issue with JSON.NET is its default handling of dates (although its been awhile since I've used it and this problem may have been rectified).

devinrader:
@abatishchev @andyday this conversation is exactly why we just need a mechanism that allows the user to override the default Json serializer.  JavaScriptSerializer will continue to be the default option since its available out of the box on all of the platforms that we want to support and does a good enough job for most users.

Regarding the comment about JSON.NET supporting portability, there are actually much bigger issues there than just JSON serialization as System.Security.Cryptography is not available as a portable library.  There is a version of JWT that I built in the portable branch of this repo if you want to take a look.  I ended up taking a dependency on PCLCrypto to replace the crypto capabilities.

devinrader:
Also @andyday I'm not sure that we should add logging directly to the JWT class.  Instead why not make the Serialize/Deserialize methods virtual and if a user wants to inject some type of logging of the serislized values, they can override them.

andyday:
this has moved to #25

PR #26, opened 2015-05-25
Add DecodeToObject<T> overload with a byte[] key

Zagitta:
A simple addition that makes the `DecodeToObject` API consistent with the regular `Decode` one.


devinrader:
Thanks for the contribution.  If we're going to add this overload to the generic DecodeToObject<T>, should we also add it to the none-generic version of the method?

Zagitta:
That's actually a very good point, I'ved added a commit which does that :)

devinrader:
Awesome.  Thanks for the contribution!

PR #33, opened 2015-10-16
Improve thrown exceptions. Add a distinct TokenExpiredException

dstj:
An expired token may need to be handled differently than a invalid signature. A distinct thrown exception makes it easier to handle that case.

Throwing "FormatException" for invalid "exp" format is more precise than throwing a generic "Exception"

### DIFF
--- a/JWT.Tests/DecodeTests.cs
+++ b/JWT.Tests/DecodeTests.cs
@@ -113,8 +113,8 @@ namespace JWT.Tests
         }
 
         [TestMethod]
-        [ExpectedException(typeof(SignatureVerificationException))]
-        public void Should_Throw_On_Expired_Token()
+        [ExpectedException(typeof(TokenExpiredException))]
+        public void Should_Throw_On_Expired_Claim()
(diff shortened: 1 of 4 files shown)


abatishchev:
It might be a breaking change to throw another type of exception but I'd go with it. Would you update the pr if this would be necessary?

dstj:
Hi @abatishchev, 

I think it's an improvement that is worth a small breaking change. How would you want me to update the PR, it's on top of master I believe.
Thanks.

abatishchev:
Just wanted to make sure you're still interested :)

SteveK-GS:
Given that token expiry previously threw a SignatureVerificationException, it might be less of a breaking change if TokenExpiredException inherited from SignatureVerificationException?

abatishchev:
Actually makes sense to me too. Would you mind to submit one more PR, please?

dstj:
Done see PR #64.
I also created PR #63 to provide more information in the TokenExpiredException

PR #37, opened 2016-03-28
Add support for nbf (Not Before) token claim validation

jhurdlow:
Added code and tests to support validation of 'nbf' claims in payload per section 4.1.5 of the JWT specification.

### DIFF
--- a/JWT/JWT.cs
+++ b/JWT/JWT.cs
@@ -173,6 +173,25 @@ namespace JWT
                     throw new SignatureVerificationException("Token has expired.");
                 }
             }
+            // verify nbf claim https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.5
+            if (payloadData.ContainsKey("nbf") && payloadData["nbf"] != null)
+            {
+                int nbf;
+                try
+                {
+                    nbf = Convert.ToInt32(payloadData["nbf"]);
+                }
+                catch (Exception)
+                {
+                    throw new SignatureVerificationException("Claim 'nbf' must be an integer.");
+                }
+
+                var secondsSinceEpoch = Math.Round((DateTime.UtcNow - UnixEpoch).TotalSeconds);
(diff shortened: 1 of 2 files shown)


JWT/JWT.cs
```diff
@@ -173,6 +173,25 @@ private static void Verify(string decodedCrypto, string decodedSignature, string
                     throw new SignatureVerificationException("Token has expired.");
                 }
             }
+            // verify nbf claim https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.5
+            if (payloadData.ContainsKey("nbf") && payloadData["nbf"] != null)
```
abatishchev: You may want to use here:
```
object nbfObj;
if (payloadData.TryGetValue(""nbf", out nbfObj) && nbfObj is int)
{
    int nbf = (int)nbfObj;
}
```

jhurdlow: Doing as you suggest would remove symmetry with the 'exp' claim validation behavior. If that was done I would expect both to be changed, and they would lose the "Claim X must be an integer" exception (unless you nested the 'is int' check).

abatishchev: Yeah, consistency makes sense. I submitted a refactoring PR but the author never cared to review it.
P.S. you still can have proper exception by splitting the if condition into 2:
```csharp
object nbfObj;
if (payloadData.TryGetValue("nbf", out nbfObj))
{
    if (!nbfObj is int)
        throw new SignatureVerificationException("Claim 'nbf' must be an integer.");
    int nbf = (int)nbfObj;
}
```

abatishchev: hi, do you still care about this pr? Would you mind to review/refactor it once again?

abatishchev:
Please rebase on the latest

abatishchev:
Can you please create a new branch based on the v2 codebase, see #67? Thanks!

abatishchev:
Redone in #81 for the v2 codebase

PR #42, opened 2016-06-18
String "keys" lead to inconsistent (possibly unintended) results:

lwg-galuise:
Consider ensuring that the methods accepting a string "key" accept only a hexadecimal string and then decode that string directly to the bytes said hexadecmial characters represent.  Or possibly consider removing the Encode/Decode methods that accept a string parameter for a key.

HMACSHA256 algorithms generally take an array of bytes as the data to be signed as well as another array of bytes as the key.

In this JWT implementation the `Encoding.UTF8.GetBytes(key)` call will simply convert the string to its UTF8 character bytes.  For example,

```csharp
Encoding.UTF8.GetBytes("0123456789abcdef") ==
    byte[]
    {
        0x30, 0x31, 0x32, 0x33,
        0x34, 0x35, 0x36, 0x37,
        0x38, 0x39, 0x61, 0x62,
        0x63, 0x64, 0x65, 0x66 // 16 bytes
    }

```

and

```csharp
Encoding.UTF8.GetBytes("0123456789ABCDEF") ==
    byte[]
    {
        0x30, 0x31, 0x32, 0x33,
        0x34, 0x35, 0x36, 0x37,
        0x38, 0x39, 0x41, 0x42,
        0x43, 0x44, 0x45, 0x46 // 16 bytes
    }
```

Where as

```csharp
"0123456789abcdef".FromHex() ==
    byte[]
    {
        0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef // only 8 bytes
    }
```

and

```csharp
"0123456789ABCDEF".FromHex() ==
    byte[]
    {
        0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef // only 8 bytes
    }
```

Not only does the `Encoding.UTF8.GetBytes()` method above lead to double the intended length byte array it also makes the hexadecimal string value case sensitive:   A hexadecimal `"a"` and a hexadecimal `"A"` both represent a decimal value of `"10"`.  However, a UTF8 character `"a"` represents a hexadecimal value of `0x61` whereas a UTF8 character character `"A"` represents a hexadecimal value of `0x41`.

### DIFF
--- a/JWT/JWT.cs
+++ b/JWT/JWT.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Security.Cryptography;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace JWT
 {
@@ -10,6 +12,30 @@ namespace JWT
         HS256,
         HS384,
         HS512
+    }
+
+    public static class Extensions
+    {
+        private static readonly Regex HEX_SANITY =
+            new Regex(@"^[0-9a-fA-F]+$");
+
+        public static byte[] FromHex(this string hex)
+        {
+            if (0 != (hex.Length % 2))
+                throw new ArgumentOutOfRangeException(
+                    "A hex string should be an even number of characters");
+            else if(!HEX_SANITY.IsMatch(hex))
+                throw new ArgumentOutOfRangeException(
+                    "Hexadecimal numbers consists only of the digits 0-9 " +
+                    "and the letters a-z");
+
+            using (MemoryStream ms = new MemoryStream())
+            {
+                for (int i = 0; i < hex.Length; i += 2)
+                    ms.WriteByte((Convert.ToByte(hex.Substring(i, 2), 16)));
+                return ms.ToArray();
+            }
+        }
     }
 
     /// <summary>
@@ -91,7 +117,7 @@ namespace JWT
         /// <returns>The generated JWT.</returns>
         public static string Encode(IDictionary<string, object> extraHeaders, object payload, string key, JwtHashAlgorithm algorithm)
         {
-            return Encode(extraHeaders, payload, Encoding.UTF8.GetBytes(key), algorithm);
+            return Encode(extraHeaders, payload, key.FromHex(), algorithm);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


abatishchev:
But what if a regular string is accepted?

lwg-galuise:
Sorry I missed this response as I haven't checked in with this project in a while.

Personally, I would say force the "application programmer" to convert their "key" to a byte[] array and remove the string method altogether.  My reasoning is that it's too easy to overlook the difference between a string and a byte array (which would seems a lot of developers confuse at first).  

If you take a look at the Microsoft's System.Security.Cryptography namespace all of the Ciphers and Hash algorithms receive their keys, initialization vectors, clear text, and cipher text as byte[] arrays.  If the developer wants to encrypt a string as UTF8 clear text said developer is forced to first pass said clear text string through the Encoding.UTF8.GetBytes() method to get a UTF8 encoding of the string:

```
// Stored internally as Unicode (16-bit) U+FFFF by .NET
string clearText = "This is my secret clear text to be encrypted";

// clearTextBytes now has a byte array in UTF8 (8-bit) encoding of the above string
byte[] clearTextBytes = Encoding.UTF8.GetBytes(clearText);
```

By not accepting a string Microsoft is kind of forcing the developer to know how the format / encoding of the data they want encrypted as well as the byte array used for the Key and Initialization vector.  

I'm working with a product where the developer doesn't seem to know the difference between a string and a hexadecimal encoded binary string.  They specified a "string" key like "0123456789abcdef0123456789abcdef" which they intended that to mean (as their documentation states):

```
byte[]
{
	0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef,
	0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef
}
```

However, since the string is being passed through Encoding.UTF8.GetBytes() the actual byte array being derived is:

```
byte[]
{
	0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
	0x38, 0x39, 0x41, 0x42, 0x43, 0x44, 0x45, 0x46,
	0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
	0x38, 0x39, 0x41, 0x42, 0x43, 0x44, 0x45, 0x46
}
```

I understand the "utility" and convenience by providing a string based method.  However since this is specifically a security / authentication library I feel this possible confusion should not exist.  I do realize this is somewhat of a breaking change (to people who may have already implemented), but I do feel strongly about the above.

abatishchev:
Thanks for the explanation! It makes sense to me now. I think as a first phase we can mark these methods obsolete, then make uncompilable.

abatishchev:
Can you please create a pr based on the v2 codebase, see #67? Thanks!

PR #45, opened 2016-08-10
Remove unnecessary ToArray()

jusfr:
as title


abatishchev:
Actually they were needed since in .NET 3.5 [there is no overload](https://msdn.microsoft.com/en-us/library/57a79xd0(v=vs.90).aspx) accepting `IEnumerable`. We both need to compile branches before submitting/accepting pull requests ;)

However, I plan to upgrade to .NET 4.5 as part of ver. 2.0.

Issue #55, opened 2016-09-16
RFC: Ideas for v. 2.0


abatishchev:
- ~~Go away from static API in favor of instance-based API with a number of separate classes~~:
  - `JwtEncoder : IJwtEncoder`
  - `JwtDecoder : IJwtDecoder`
  - `JwtValidator : IJwtValidator`
- ~~Mark static API `[Obsolete]`~~
- ~~Make static API a wrapper on top of instance-based API~~
- Remove static API in the future versions (such as 2.1 or 3.0)

abatishchev:
* Replace MSTest with xUnit

abatishchev:
* Make v2 portable or .NET Core compatible or both
 - Community contribution would be needed, I have no experience here (yet)

abatishchev:
* Rebrand *Jwt implementation for .NET* as *Jwt.Net*
 - This repo always belongs to the jwt-dotnet organization but now do this officially, i.e. everywhere including readme and nuget package's name and description
 - Other options: Jwt.NET, JWT.NET, JWT.Net

abatishchev:
* Inject `IJsonSerializer` into ctors of `JwtEncoder` and `JwtDecoder`
 - Or split it into two: `IJsonSerializer` and `IJsonDeserializer`
* Rename it to `IJwtSerializer` to have consistent naming?

abatishchev:
* Set `[AssemblyVersion]` to 2.0.0.0 and never change

abatishchev:
* Replace `JavascriptSerializer` with Json.Net

alexfoxgill:
A suggestion:

* Remove the direct dependency on DateTime.UtcNow

abatishchev:
Can you please explain why, what is the intend? Currently it calculates the difference between UcNow and the start of UNIX epoch fot token expiration.

alexfoxgill:
Mainly for the ability to mock while testing:

* http://stackoverflow.com/questions/2425721/unit-testing-datetime-now/2425739#2425739
* https://blogs.msdn.microsoft.com/ploeh/2007/05/12/testing-against-the-current-time/

There may be other uses but in general it's a good idea to use dependency injection for things that can't be mocked, or at least to provide some method for doing this if it's not the default

abatishchev:
Oh, right, Of course. I thought maybe you just have some other thoughts/needs. I usually do this:

```csharp
public interface IDateTimeProvider
{
	DateTime GetNow();
}
[...]

PR #60, opened 2016-09-29
Convert JWT library to dotnetcore

alistair:
I believe this is now finished but would like a good review of the changes I have made.   I don't have VS and have instead been using vim/project rider on linux to work on this so please test on windows with VS.

This pull request converts the project to dotnetcore.  It produces a nuget package targeting .NET 3.5 and NETSTANDARD1_3 ( which corresponds to .NET 4.5.1 i believe )

Unit tests are only run against the NETSTANDARD1_3 version of code as I was unable to get the other version to run at all.  This seems like an outstanding bug with dotnet.


abatishchev:
Great! Thanks for your work and thanks for sharing it! I'm absolutely support switching from any other Json implementation to Json.Net.

Please also see https://github.com/jwt-dotnet/jwt/issues/55. I logged this idea there.

alistair:
@abatishchev I consider this ready now.  Do you?

JWT.sln
```diff
@@ -3,30 +3,25 @@ Microsoft Visual Studio Solution File, Format Version 12.00
...
 VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JWT", "src\JWT\JWT.csproj", "{A80B51B8-DDF6-4026-98A4-B59653E50B38}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{513CE2B5-E0D6-43BC-998A-A02CB2875479}"
+ProjectSection(SolutionItems) = preProject
+	global.json = global.json
+EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JWT.Tests", "tests\JWT.Tests\JWT.Tests.csproj", "{BF568781-D576-4545-A552-4DC839B1AF14}"
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "JWT", "src\JWT\JWT.xproj", "{00000000-0000-0000-0000-000000000000}"
```
abatishchev: Do you know why it's empty guid now?

src/JWT/JWT.cs
```diff
@@ -15,7 +15,11 @@ public static class JsonWebToken
         /// <summary>
         /// Pluggable JSON Serializer
         /// </summary>
+#if !NETSTANDARD1_3
```
abatishchev: I think we should jump 2.0 and make this change for either platforms

tests/JWT.Tests/DecodeTests.cs
```diff
@@ -1,14 +1,16 @@
...
 using System.Collections.Generic;
+
+#if !NETCOREAPP1_0
 using System.Web.Script.Serialization;
+#endif
 
 using FluentAssertions;
 
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Xunit;
```
abatishchev: So you're switching over Xunit?

tests/JWT.Tests/EncodeTests.cs
```diff
@@ -1,72 +1,73 @@
...
+using Xunit;
 
 namespace JWT.Tests
 {
-    [TestClass]
     public class EncodeTests
     {
         private static readonly Customer _customer = new Customer { FirstName = "Bob", Age = 37 };
 
         private const string _token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJGaXJzdE5hbWUiOiJCb2IiLCJBZ2UiOjM3fQ.cr0xw8c_HKzhFBMQrseSPGoJ0NPlRp_3BKzP96jwBdY";
```
abatishchev: Shouldn't then we use these variables as test's parameters in conjunction with `[Theory]`?

tests/JWT.Tests/project.json
```diff
@@ -0,0 +1,38 @@
...
+    "version": "1.0.0-*",
+    "testRunner": "xunit",
+    "dependencies": {
+        "dotnet-test-xunit": "2.2.0-preview2-build1029",
+        "xunit": "2.2.0-beta2-build3300",
+        "FluentAssertions": "4.14.0",
+        "JWT": "1.3.7-*"
+    },
+    "frameworks": {
+        /*"net451": {
```
abatishchev: I think this should be 3.5 but why commented?

tests/JWT.Tests/DecodeTests.cs
alistair: yes.  xunit is like the default dotnetcore unit testing framework.   All the examples Ive seen use it and I wouldn't have a clue how to setup anything else :)

tests/JWT.Tests/project.json
alistair: xunit only supports >= net451.   I commented this because I was getting errors when attempting to run the tests.   I will try track down the dotnet issue I was reading about this which might give some more context.

tests/JWT.Tests/EncodeTests.cs
alistair: Probably,  I was mostly concerned with making the most minimal of changes required to get this working.

src/JWT/JWT.cs
alistair: Im assuming you mean BOTH platforms.   I will make this change then.

JWT.sln
alistair: I will have to double check.  I updated the solution using project rider so im not 100% confident it is correct.   I don't actually have VS.

abatishchev: I'll try to compile in VS2015 once get access to my dev vm

src/JWT/JWT.cs
abatishchev: Yeah, right

tests/JWT.Tests/DecodeTests.cs
abatishchev: Love this! Just wanted to make sure, was going to switch over Xunit anyway with v2

tests/JWT.Tests/EncodeTests.cs
abatishchev: Sure, makes sense, this would be intend too. Let's keep it as is, not a big deal, not a big difference either.

tests/JWT.Tests/project.json
abatishchev: This might be another reason to bump the version to 2.0 since is a breaking change. Personally I don't mind at all.

abatishchev:
hey @alistair, i'm sorry it ran out of my scope. how do you think is everything ready to be merged?

abatishchev:
Hi, I made all changes I was going to bring in v2 in this PR: https://github.com/jwt-dotnet/jwt/pull/67. If you could take a look, maybe apply your changes on top of it, so we could bring the .NET Core compatibility as part of v2?

abatishchev:
I merge v2 into master. Anybody, please update the pr/rebase it.

nbarbettini:
Since this PR was using the old (project.json-based) tooling, I can build a new PR for .NET Core support. Let me know if that works for you @abatishchev!

abatishchev:
This would be perfect, thanks!

paddyfink:
Hey guys, do you have any release date for this feature please?

abatishchev:
Let's move the discussion to #72

PR #63, opened 2016-12-06
Add payloadData and expiration to the TokenExpiredException exception

dstj:
Adding more contextual information to the `TokenExpiredException` for usage by the parent application

### DIFF
--- a/src/JWT/JWT.cs
+++ b/src/JWT/JWT.cs
@@ -264,12 +264,11 @@ namespace JWT
             var secondsSinceEpoch = Math.Round((DateTime.UtcNow - UnixEpoch).TotalSeconds);
             if (secondsSinceEpoch >= expInt)
             {
-               var tokenExpiredException = new TokenExpiredException("Token has expired.")
-               {
-                  Expiration = UnixEpoch.AddSeconds(expInt),
-                  PayloadData = payloadData
-               };
-               throw tokenExpiredException;
+                throw new TokenExpiredException("Token has expired.")
+                {
+                    Expiration = UnixEpoch.AddSeconds(expInt),
+                    PayloadData = payloadData
+                };
             }
         }
 
--- a/src/JWT/JWT.cs
+++ b/src/JWT/JWT.cs
@@ -264,10 +264,12 @@ namespace JWT
             var secondsSinceEpoch = Math.Round((DateTime.UtcNow - UnixEpoch).TotalSeconds);
             if (secondsSinceEpoch >= expInt)
             {
-                var tokenExpiredException = new TokenExpiredException("Token has expired.");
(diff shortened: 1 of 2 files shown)


src/JWT/TokenExpiredException.cs
```diff
@@ -1,12 +1,38 @@
...
+       public TokenExpiredException(string message)
             : base(message)
         {
         }
+
+        public Dictionary<string, object> PayloadData
+        {
+            get
+            {
+               if (Data.Contains(PayloadDataKey))
```
abatishchev: Can you move this snippet to a helper method? To reuse, also make both getters one-liners. Also do we really need to check a key to exist? I'd let it throw.

src/JWT/JWT.cs
```diff
@@ -264,7 +264,12 @@ public static void Verify(string payloadJson, string decodedCrypto, string decod
             var secondsSinceEpoch = Math.Round((DateTime.UtcNow - UnixEpoch).TotalSeconds);
             if (secondsSinceEpoch >= expInt)
             {
-                throw new TokenExpiredException("Token has expired.");
+               var tokenExpiredException = new TokenExpiredException("Token has expired.")
```
abatishchev: You can throw right away, without a variable assignment.

abatishchev:
Thanks for this change! I left few minor comments.

src/JWT/TokenExpiredException.cs
dstj: I was debating whether to throw or not... The problem I have with throwing is that since you requested that `TokenExpiredException` inherits from `SignationVerificationException` (see PR #33), the `Received` and `Expected` properties will not be assigned for `TokenExpiredException`... so making a `catch (SignationVerificationException ex)` may crash if we don't do that...

src/JWT/JWT.cs
abatishchev: Cool! Can you please format the document (usually Ctrl+K,D) to make sure no extra/missing spaces were added/removed?

src/JWT/TokenExpiredException.cs
abatishchev: Sure, let's keep it safe. Can you please move it a helper? And remove a space inside the cast operator, to follow the common style.

src/JWT/JWT.cs
dstj: I'll try. I'll need to change my settings, cause Ctrl-K,D will result in tabs all over otherwise! 😛

abatishchev: Yeah, I feel you pain, I'm a tabs guy too! Don't get me started ;) Usually what I do is format then find and replace \t with "    ".

dstj: Hehe ;)
PR updated

PR #64, opened 2016-12-06
TokenExpiredException inherits from SignatureVerificationException + Refactor 

dstj:
... and Refactor SignatureVerificationException to use properties instead of the Data collection directly

### DIFF
--- a/src/JWT/JWT.cs
+++ b/src/JWT/JWT.cs
@@ -239,10 +239,12 @@ namespace JWT
         {
             if (decodedCrypto != decodedSignature)
             {
-	            var signatureVerificationException = new SignatureVerificationException("Invalid signature");
-				signatureVerificationException.Data.Add("Expected", decodedCrypto);
-				signatureVerificationException.Data.Add("Received", decodedSignature);
-	            throw signatureVerificationException;
+               var signatureVerificationException = new SignatureVerificationException("Invalid signature")
+               {
+                  Expected = decodedCrypto,
+                  Received = decodedSignature
+               };
+               throw signatureVerificationException;
             }
 
             // verify exp claim https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.4
--- a/src/JWT/JWT.cs
+++ b/src/JWT/JWT.cs
@@ -240,10 +240,10 @@ namespace JWT
             if (decodedCrypto != decodedSignature)
             {
                 throw new SignatureVerificationException("Invalid signature")
-               {
-                   Expected = decodedCrypto,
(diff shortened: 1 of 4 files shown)


src/JWT/JWT.cs
```diff
@@ -239,10 +239,12 @@ public static void Verify(string payloadJson, string decodedCrypto, string decod
         {
             if (decodedCrypto != decodedSignature)
             {
-	            var signatureVerificationException = new SignatureVerificationException("Invalid signature");
-				signatureVerificationException.Data.Add("Expected", decodedCrypto);
-				signatureVerificationException.Data.Add("Received", decodedSignature);
-	            throw signatureVerificationException;
+               var signatureVerificationException = new SignatureVerificationException("Invalid signature")
```
abatishchev: Can you please to make here the same change as in #63?

src/JWT/SignatureVerificationException.cs
```diff
@@ -4,9 +4,34 @@ namespace JWT
...
         public SignatureVerificationException(string message)
             : base(message)
         {
         }
+
+        public string Expected
+        {
+            get
+            {
+               if (Data.Contains(ExpectedKey))
```
abatishchev: And here, please.

src/JWT/JWT.cs
dstj: I'll rebase on PR #63 because I'll move the helper method in the `SignationVerificationException`...

dstj: Done, I forced pushed my PR, we'll see how github likes it...

abatishchev:
btw wouldn't it easier to have 1 PR for all these changes? and close the another one. Just thinking out loud.

src/JWT/JWT.cs
abatishchev: I think that one good enough, especially if we'll reside the helper method here. Pelsse go ahead and rebase, I'll merge

dstj:
@abatishchev, I've rebased the PR. Yes it would have been easier (for me anyway) to do only one PR, but I wanted to keep the commit separated as they are not the same intention....

abatishchev:
Cool! Would you mind please to format all documents, and then I'll merge right away.

abatishchev:
@dstj please format and I'll merge and publish

dstj:
@abatishchev Sorry about the delay, it is done. A few spaces were added...

abatishchev:
Thanks again for your contributions! I need to publish this now

abatishchev:
These change have been released as 1.4.1-beta (I'm bit scared to remove the per-release marker). Thanks again!

PR #81, opened 2017-04-03
Add support for nbf (Not Before) token claim validation in V2

abatishchev:
Redoing https://github.com/jwt-dotnet/jwt/pull/37


abatishchev:
@jhurdlow, @nbarbettini please take a look

jhurdlow:
Looks like the 'nbf' will never be checked if 'exp' doesn't exist. 'nbf' should still be checked in that case.

abatishchev:
Yes, I found that too (tests failed) so I reworked that part.

abatishchev:
@jhurdlow can you please take a look once again?

abatishchev:
Pushed to NuGet as [2.3.0-beta](https://www.nuget.org/packages/JWT/2.3.0-beta). Thanks you both!

PR #82, opened 2017-04-03
Moving Base64UrlEncode/Base64UrlDecode to separate entity

abatishchev:
- Adding IBase64UrlEncoder implemented by JwtBase64UrlEncoder. Calling it from JsonWebToken.Base64UrlEncode/Base64UrlDecode.
- Updating tests

### DIFF
--- a/package.cmd
+++ b/package.cmd
@@ -3,9 +3,9 @@
 if not exist package mkdir package
 if not exist package mkdir package
 if not exist package\lib mkdir package\lib
-if not exist package\lib\net35 mkdir package\lib\net35
+if not exist package\lib\net45 mkdir package\lib\net45
 
 msbuild src\JWT.sln -p:Configuration=Release
-copy src\JWT\bin\Release\JWT.dll package\lib\net35
+copy src\JWT\bin\Release\JWT.dll package\lib\net45
 
 tools\nuget.exe pack JWT.nuspec -BasePath package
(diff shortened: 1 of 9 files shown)


abatishchev:
I'll wait CI/CD to run tests automatically first

package.cmd
```diff
@@ -1,11 +1,11 @@
...
-
+﻿tools\nuget.exe update -self
+
+if not exist package mkdir package
+if not exist package mkdir package
+if not exist package\lib mkdir package\lib
+if not exist package\lib\net35 mkdir package\lib\net35
+
+msbuild src\JWT.sln -p:Configuration=Release
+copy src\JWT\bin\Release\JWT.dll package\lib\net35
```
nbarbettini: Looks like some extra changes got moved into this PR?

src/JWT/IBase64UrlEncoder.cs
```diff
@@ -0,0 +1,9 @@
+namespace JWT
+{
+    public interface IBase64UrlEncoder
+    {
+        string UrlEncode(byte[] input);
```
nbarbettini: Seems a bit redundant to have `Url` twice, perhaps the method could just be `Encode` and `Decode`?

abatishchev: Yeah, was thinking about this too. Drop the prefix.

Issue #75, opened 2017-03-28
Update to JSON.NET 10


nbarbettini:
There's a new major version of JSON.NET: https://www.nuget.org/packages/Newtonsoft.Json/10.0.1

Should this JWT >2.x depend on this? I tested it briefly and there aren't any problematic API changes.

abatishchev:
I would build against the latest but do not limit dependency version in nuget since Json.Net is usually backward-compatible down to 6.x or even 4.x.

nbarbettini:
I will do some testing and figure out how to widen the dependency version in the package 👍

abatishchev:
Done in #83.

PR #83, opened 2017-04-08
Updating Json.Net to v10, other dev/test dependencies

abatishchev:
Based on #75 discussion.

The most important change: **removed the beta prefix**.

### DIFF
--- a/JWT.nuspec
+++ b/JWT.nuspec
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>JWT</id>
-    <version>2.3.0-beta</version>
+    <version>2.3.1</version>
     <authors>John Sheehan, Michael Lehenbauer, Alexander Batishchev</authors>
     <owners>johnsheehan, devinrader, abatishchev</owners>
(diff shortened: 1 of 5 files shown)


nbarbettini:
LGTM 👍

ericbock:
It seems like the [.nuspec](https://github.com/jwt-dotnet/jwt/blob/v2/JWT.nuspec#L15) in v2 needs to be updated to account for the newer version of Json.Net. If I use any version of Json.Net prior to 10.0.1 I get: Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0

abatishchev:
Versions of Jon.Net from 6x till 10.x are binary backward compatible. So I don't think we need to update the nuspec. Instead please try to add adding assembly binding redirect into your Web/App.config, e.g. from version 0.0.0.0 till 10.0.0.0 to your version, i.e. 6.0.0.0 or 10.0.0.0, etc.

ericbock:
That worked, thanks.

PR #89, opened 2017-04-20
Update exp and nbf validation to use double parsing

zemien:
Update exp and nbf validation to be based on strictly-parsed doubles for 2 reasons: Support dates beyond year 2038, and prevent treating nulls as 0.

### DIFF
--- a/src/JWT/JwtValidator.cs
+++ b/src/JWT/JwtValidator.cs
@@ -32,7 +32,6 @@ namespace JWT
 
             var now = _dateTimeProvider.GetNow();
             var secondsSinceEpoch = Math.Round((now - UnixEpoch).TotalSeconds);
-            const string claimMustBeDoubleFormat = "Claim '{0}' must be a double.";
 
             // verify exp claim https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.4
             object expObj;
@@ -40,7 +39,7 @@ namespace JWT
             {
                 if (expObj == null)
                 {
-                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "exp"));
+                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
                 }
 
                 double expInt;
@@ -50,7 +49,7 @@ namespace JWT
                 }
                 catch
                 {
-                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "exp"));
+                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
                 }
 
                 if (secondsSinceEpoch >= expInt)
@@ -69,7 +68,7 @@ namespace JWT
             {
                 if (nbfObj == null)
                 {
-                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "nbf"));
+                    throw new SignatureVerificationException("Claim 'nbf' must be a double.");
                 }
 
                 double nbfInt;
@@ -79,7 +78,7 @@ namespace JWT
                 }
                 catch
                 {
-                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "nbf"));
+                    throw new SignatureVerificationException("Claim 'nbf' must be a double.");
                 }
 
                 if (secondsSinceEpoch < nbfInt)
--- a/src/JWT/JwtValidator.cs
+++ b/src/JWT/JwtValidator.cs
@@ -32,6 +32,7 @@ namespace JWT
 
             var now = _dateTimeProvider.GetNow();
             var secondsSinceEpoch = Math.Round((now - UnixEpoch).TotalSeconds);
+            const string claimMustBeDoubleFormat = "Claim '{0}' must be a double.";
 
             // verify exp claim https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.4
             object expObj;
@@ -39,7 +40,7 @@ namespace JWT
             {
                 if (expObj == null)
                 {
-                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
+                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "exp"));
                 }
 
                 double expInt;
@@ -49,7 +50,7 @@ namespace JWT
                 }
                 catch
                 {
-                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
+                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "exp"));
                 }
 
                 if (secondsSinceEpoch >= expInt)
@@ -68,7 +69,7 @@ namespace JWT
             {
                 if (nbfObj == null)
                 {
-                    throw new SignatureVerificationException("Claim 'nbf' must be a double.");
+                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "nbf"));
                 }
 
                 double nbfInt;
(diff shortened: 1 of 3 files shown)


zemien:
#89 fix

src/JWT/JwtValidator.cs
```diff
@@ -37,14 +37,14 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
             object expObj;
             if (payloadData.TryGetValue("exp", out expObj))
             {
-                int expInt;
+                double expInt;
                 try
                 {
-                    expInt = Convert.ToInt32(expObj);
+                    expInt = double.Parse(expObj.ToString());
```
abatishchev: `Double.ToString()` has current culture issues. Please use `Convert.ToDouble()` instead

src/JWT/JwtValidator.cs
```diff
@@ -61,14 +61,14 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
             object nbfObj;
             if (payloadData.TryGetValue("nbf", out nbfObj))
             {
-                int nbfInt;
+                double nbfInt;
                 try
                 {
-                    nbfInt = Convert.ToInt32(nbfObj);
+                    nbfInt = double.Parse(nbfObj.ToString());
```
abatishchev: And here please

tests/JWT.Tests.Core/DecodeTests.cs
```diff
@@ -79,14 +89,42 @@ public void DecodeToObject_Should_Throw_Exception_On_Expired_Claim()
         }
 
         [Fact]
+        public void DecodeToObject_Should_DecodeToken_On_Exp_Claim_After_Year2038()
+        {
+            var serializer = new JsonNetSerializer();
+            JsonWebToken.JsonSerializer = serializer;
+
+            var post2038 = new DateTime(3000, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
abatishchev: Should we use 2038/1/1 + 1 second here? To justify variable's name

tests/JWT.Tests.Core/DecodeTests.cs
```diff
@@ -79,14 +89,42 @@ public void DecodeToObject_Should_Throw_Exception_On_Expired_Claim()
...
+            var actualPayload = JsonWebToken.Decode(validToken, "ABC", true);
+
+            actualPayload.Should().Be(expectedPayload);
+        }
+        
+        [Fact]
         public void DecodeToObject_Should_Throw_Exception_Before_NotBefore_Becomes_Valid()
         {
-            var nbf = (int)(DateTime.UtcNow.AddHours(1) - JwtValidator.UnixEpoch).TotalSeconds;
+            var post2038 = new DateTime(3000, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
abatishchev: And here

zemien:
I've reverted back to using Convert and added a separate null check to ensure it doesn't get converted to 0.

src/JWT/JwtValidator.cs
```diff
@@ -32,19 +32,25 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
...
+                double expInt;
                 try
                 {
-                    expInt = Convert.ToInt32(expObj);
+                    expInt = Convert.ToDouble(expObj);
                 }
                 catch
                 {
-                    throw new SignatureVerificationException("Claim 'exp' must be an integer.");
+                    throw new SignatureVerificationException(string.Format(claimMustBeDoubleFormat, "exp"));
```
abatishchev: Let's don't extract strings into const variables to keep the source code smaller and more readable

tests/JWT.Tests.Core/DecodeTests.cs
```diff
@@ -67,10 +67,20 @@ public void DecodeToObject_Should_Throw_Exception_On_Invalid_Expiration_Claim()
         }
 
         [Fact]
+        public void DecodeToObject_Should_Throw_Exception_On_Null_Expiration_Claim()
```
abatishchev: Sorry for confusing. Encode/Decode tests are for now-obsolete static API which we'll remove shortly. Can you please move new tests to JwtEncoder/Decoder tests?

zemien:
Reverted the constant change. I knew it would be contentious so made it a separate commit ;)
Ported the tests to JwtDecoder

abatishchev:
Looks good to me.  Thanks for this work!

Let's keep till EoD so @nbarbettini could give a quick look. And I'll merge.

Thanks again :)

nbarbettini:
Thanks for the Y2038 catch, @zemien! One question - why `double` and not `long` (Int64)?

zemien:
Hmm good question. I suppose most approaches for calculating epoch in .Net uses the TotalSeconds property which is a double. Logically "seconds since epoch" do not appear to represent fractional seconds, but then again is there a reason to stop someone from specifying 0.5 (Jan 1, 1970 at 00:00:00.500)?

src/JWT/JwtValidator.cs
```diff
@@ -37,14 +37,19 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
...
+                double expInt;
                 try
                 {
-                    expInt = Convert.ToInt32(expObj);
+                    expInt = Convert.ToDouble(expObj);
                 }
                 catch
                 {
-                    throw new SignatureVerificationException("Claim 'exp' must be an integer.");
+                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
```
nbarbettini: I think "Claim 'exp' must be a number." would remove some potential confusion from the error message.

src/JWT/JwtValidator.cs
```diff
@@ -61,14 +66,19 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
             object nbfObj;
             if (payloadData.TryGetValue("nbf", out nbfObj))
             {
-                int nbfInt;
+                if (nbfObj == null)
+                {
+                    throw new SignatureVerificationException("Claim 'nbf' must be a double.");
```
nbarbettini: Ditto

src/JWT/JwtValidator.cs
```diff
@@ -61,14 +66,19 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
...
+                double nbfInt;
                 try
                 {
-                    nbfInt = Convert.ToInt32(nbfObj);
+                    nbfInt = Convert.ToDouble(nbfObj);
                 }
                 catch
                 {
-                    throw new SignatureVerificationException("Claim 'nbf' must be an integer.");
+                    throw new SignatureVerificationException("Claim 'nbf' must be a double.");
```
nbarbettini: Ditto3

src/JWT/JwtValidator.cs
```diff
@@ -37,14 +37,19 @@ public void Validate(string payloadJson, string decodedCrypto, string decodedSig
             object expObj;
             if (payloadData.TryGetValue("exp", out expObj))
             {
-                int expInt;
+                if (expObj == null)
+                {
+                    throw new SignatureVerificationException("Claim 'exp' must be a double.");
+                }
+
+                double expInt;
```
nbarbettini: Nitpick: This should be `expValue` or just `exp`.

nbarbettini:
> is there a reason to stop someone from specifying 0.5

No reason, I think using double is fine. I forgot that `TotalSeconds` was a floating-point value.

Code looks good! 👍 Sent you a few minor comments.

zemien:
Done, please review

Issue #95, opened 2017-04-26
AppVeryor picks-up no tests


abatishchev:
Log says:

>Executing task: Test
>Running 0 tests
>Finished executing task: Test

Can you please take a look?

nbarbettini:
Good catch! Turns out there were some issues with XUnit. It was trickier than I thought to get both the .NET Standard and .NET Framework tests working. Should be fixed in #97.

PR #97, opened 2017-04-27
Fix tests for both platforms

nbarbettini:
Closes https://github.com/jwt-dotnet/jwt/issues/95


abatishchev:
Doh. Was you able to get it working? If yes, can you please submit a pr?

blgrnboy:
I was able to get it working by uninstalling and reinstalling the packages, which modified the path. I will submit a PR but it may be a few days since I'm travelling.

PR #101, opened 2017-04-28
Adding IAlgorithmFactory

abatishchev:
- Adding IAlgorithmFactory
- Using IAlgorithmFactory in JwtDecoder
- Renaming AlgorithmFactory to HMACSHAAlgorithmFactory
- Adding RSAlgorithmFactory
- Adding DotSettings file


src/JWT/Algorithms/RSAlgorithmFactory.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace JWT.Algorithms
+{
+    public sealed class RSAlgorithmFactory : HMACSHAAlgorithmFactory
+    {
+        private readonly Func<X509Certificate2> _certFactory;
+
+        public RSAlgorithmFactory(Func<X509Certificate2> certFactory)
```
djavier: +1 passing a factory as a function to create the certFactory. This might turn out to be a generic `RSAlgorithmFactory` and not only rely on `X509Certificate2` if any other `RSAlgorithms` should be implemented.

abatishchev: Right. Since there will be an interface there now, it's up to consumers to implemented the way they need it. Since currently there is only 1 RS* algorithm and it requires `X509Certificate2` to function, I'll keep `RSAlgorithmFactory` title coupled with `RS256Algorithm`. Once there will be more, we can make the factory more generic. Makes sense?

djavier: Totally agree, there is no need to make it more complex, even though there are many RS signatures. Maybe i submit the pull request for the other ones ;). I'm getting hyped.

abatishchev:
Release to NuGet as [2.4.0](https://www.nuget.org/packages/JWT/2.4.0), also will merge to master

PR #103, opened 2017-05-01
Removing static API

abatishchev:
- Removing static API, tests


nbarbettini:
Nice! 🎉

nbarbettini:
FYI - I rebased this onto master and created a new PR at #107.

tests/JWT.Tests.NETFramework/DecodeTests.cs
```diff
@@ -1,78 +0,0 @@
...
-
-namespace JWT.Tests.NETFramework
-{
-    public class DecodeTests
-    {
-        [Fact]
-        public void Decode_Should_Decode_Token_To_Json_Encoded_String_With_WebScript_Serializer()
-        {
-            var serializer = new WebScriptJsonSerializer();
-            JsonWebToken.JsonSerializer = serializer;
```
nbarbettini: These tests are now missing. I think we should modify them to use an instance instead of the static `JsonWebToken` object.

tests/JWT.Tests.NETFramework/EncodeTests.cs
```diff
@@ -1,54 +0,0 @@
...
-using Xunit;
-
-namespace JWT.Tests.NETFramework
-{
-    public class EncodeTests
-    {
-        [Fact]
-        public void Should_Encode_Type_With_WebScript_Serializer()
-        {
-            JsonWebToken.JsonSerializer = new WebScriptJsonSerializer();
```
nbarbettini: Ditto

Issue #76, opened 2017-03-28
Split NuGet package into few child packages to ease dependencies, lower minimum target framework


abatishchev:
The idea is to extract few child packages:

- JwtNet.Json.NewtonSoft for JsonNet based implementation of Json (de)serialization
- JwtNet.Json.ServiceStack for ServiceStack.Text based implementations
- JwtNet.Cryptography.RS256 for System.Security.Cryptography based implementation of RS256Algorithm

nbarbettini:
I like this idea, because it will mean that the "base" functionality can exist in a package that supports the widest number of platforms/devices. The only thing to figure out is how to structure the package graph to be both flexible and have easy discoverability.

For example, if you do `install-package JWT`, should you get `JWT.Json.Newtonsoft` automatically? It would be less confusing if you did, but if it's declared as a dependency of `JWT`, you won't be able to remove it.

One solution I've used in the past is to push all of the interfaces (and maybe base functionality) into a package called `JWT.Abstractions` or `JWT.Core`, and then make the top-level `JWT` package just be an empty metapackage that serves as an easy target for discoverability. That would look like:

```
JWT (empty package, just dependencies)
  - depends on JWT.Abstractions
  - depends on JWT.Json.Newtonsoft
    - depends on JWT.Abtractions
```

So, `install-package JWT` gives you the default packages. If you want to customize the installation, you can skip `JWT` and install `JWT.Abtractions` directly plus whatever supporting packages you want.

Any thoughts?

abatishchev:
Sounds like a plan! I would just avoid calling it Core to avoid confusion with .NET Core. At work I usually call such project Contracts but we can make up something different.

Should we bump major version to 3.x since it'll be one more big breaking change.

nbarbettini:
> I would just avoid calling it Core to avoid confusion with .NET Core

Agreed 😄 

> Should we bump major version to 3.x since it'll be one more big breaking change.

I'm fine with that, would you want to fully drop the old static interface too?

abatishchev:
Yes, I think that would be the right time to do it.

zemien:
This looks like a great, easy-to-use library. I notice @abatishchev has started porting it to .Net Standard. Is there anything I can help with that process? I need something similar for my .Net Core project and I would rather help get this production-ready than do my own personal fork.

abatishchev:
hi @zemien, the port is done (committed but not published yet). hey @nbarbettini, can you please pack and push to nuget?

[...]

PR #107, opened 2017-05-13
Removing static API

nbarbettini:
Rebased #103


nbarbettini:
@abatishchev I noticed you removed these files: https://github.com/jwt-dotnet/jwt/blob/master/tests/JWT.Tests.NETFramework/DecodeTests.cs

These tests aren't duplicated anywhere else. I think maybe we should change them to using the instance (not static), instead of deleting them?

abatishchev:
So did I :-D Never mind, closed that PR in favor of this one.

Regarding the tests. IIRC, new tests are just clones of old tests except the part where static factory is assigned to substitute JSON (de)serializer. So we should be fine by deleting them. An additional review by separate pair of eyes would be helpful though!

abatishchev:
Oops, closed wrong one. Reopened.

nbarbettini:
Re: tests - I agree that the duplicate tests in `JWT.Tests.Core` are unnecessary and can be deleted. However, the tests in `JWT.Tests.NETFramework` were testing the other deserializer plugins and may still be useful. Do you want to keep the System.Web and ServiceStack tests around?

abatishchev:
Frankly I don't. This functionality goes away with the removal of static API.
To add it back and this time do it properly, there should be 2 additional/separate nuget packages (as per #76) with their own tests. Those might be just clones of standard tests but targeted against their own implementation.
Makes sense?

nbarbettini:
Ah, now I'm understanding you. Yep! Let's merge.

abatishchev:
Pushed to NuGet as [3.0.0-beta3](https://www.nuget.org/packages/JWT/3.0.0-beta3).

@nbarbettini Thanks for your help! As always :)

PR #100, opened 2017-04-28
Fix tests nuget references

blgrnboy:
Fixed nuget references

PR #109, opened 2017-05-13
Fix test builds

### LINKED ISSUE #108: Can't build after cloning new project
I reproduced the issue @blgrnboy was experiencing in #100. The package path needs to be reverted or fixed and also needs to work on the build server.

nbarbettini:
This PR contains a couple of changes:

* Removes the System.Web and ServiceStack serializers that were previously used for tests. (Deprecated in v3, replaced by #76 eventually)
* Adds some duplicate tests to JWT.Tests.NETFramework, to sanity check that the library works from a .NET Framework (not .NET Core) project. It should always work, but tests are nice. 😄 
* Fixes the package hint path issues from #100. The solution should now build fine locally and on the CI server.

Resolves #108


nbarbettini:
Good call. I forgot that it's possible to reference code files in another project. The latest change looks like it works on the build server.

nbarbettini:
I tried that at first, but xUnit wouldn't discover the tests in the project that used links.

abatishchev:
That's odd. Let me try locally, I need to see with my own eyes to fully believe :) Because it shouldn't care since suppose to scan output assembly, not input project.

abatishchev:
This is odd but by due to another reason: AppVeyor builds time out on package restore step now.

abatishchev:
Never mind, seems to work fine now

blgrnboy:
Are you guys going to merge here?

PR #118, opened 2017-05-18
Modify RS256Algorithm to be xplat friendly

### LINKED ISSUE #117: RSACryptoServiceProvider not available on Linux and Mac
When using the library from a CoreCLR project, `_cert.GetRSAPrivateKey();` can return things other than `RSACryptoServiceProvider`.

For example, when trying to use `RS256Algorithm`, I get:
```
An unhandled exception has occurred: Unable to cast object of type 'System.Security.Cryptography.RSAOpenSsl' to type 'System.Security.Cryptography.RSACryptoServiceProvider'.
```

I was going to submit a Pull Request to make everything just use the base class, RSA, but I am unsure what [this](https://github.com/jwt-dotnet/jwt/blob/d51a8a3b15b0845791479a26b5eafdb833105cfb/src/JWT/Algorithms/RS256Algorithm.cs#L34-L39) is for or why it's important.

I'd love to help make this wonderful for non-Windows users. Let me know what's up.

Thanks,
Zak

ubercow:
Use the base class `RSA` for `RS256Algorithm`, which allows use on platforms like macOS and Linux, which won't have `RSACryptoServiceProvider`.

This is a very much "works on my machine" build and probably shouldn't be merged as-is. I notably have not tested this on net35.

Particularly, further research is needed as to why the `CspParameters` and creating a new `RSACryptoServiceProvider` was needed in the old code. I don't know anything about Windows crypto, so I'm not much help here.

Fixes #117

### DIFF
--- a/src/JWT/Algorithms/RS256Algorithm.cs
+++ b/src/JWT/Algorithms/RS256Algorithm.cs
@@ -27,17 +27,12 @@ namespace JWT.Algorithms
         public byte[] Sign(byte[] key, byte[] bytesToSign)
         {
 #if NETSTANDARD1_3
-            var rsa = (RSACryptoServiceProvider)_cert.GetRSAPrivateKey();
+            var rsa = _cert.GetRSAPrivateKey();
 #else
-            var rsa = (RSACryptoServiceProvider)_cert.PrivateKey;
+            var rsa = _cert.PrivateKey;
 #endif
-            var param = new CspParameters
-            {
-                KeyContainerName = rsa.CspKeyContainerInfo.KeyContainerName,
-                KeyNumber = rsa.CspKeyContainerInfo.KeyNumber == KeyNumber.Exchange ? 1 : 2
-            };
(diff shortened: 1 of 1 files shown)


nbarbettini:
Code looks good to me at first glance. I also don't know why that other code was there. I haven't done much RSA256 crypto myself.

abatishchev:
We need few basic (smoke) tests to proceed.

Meanwhile is it possible somehow test both platforms implementation? Or AppVeyor will compile and run one or another but not both?

abatishchev:
btw by some reason one of AppVeyor checks has stuck and hasn't changed since then

nbarbettini:
I believe it's possible to build on multiple platforms using AppVeyor's build matrix feature, although I haven't tried it.

abatishchev:
It seems here's what we need: [Build configuration#Build matrix](https://www.appveyor.com/docs/build-configuration/#build-matrix)

Issue #119, opened 2017-06-22
Allow a custom Json serialization option when creating a jwt using Newtonsoft.Json


PureKrome:
Hi 👋 

Would you folks accept a PR for allowed an optional `JsonSerializer` to be added to the [JsonNetSerializer class](https://github.com/jwt-dotnet/jwt/blob/master/src/JWT/Serializers/JsonNetSerializer.cs)?

So I've got some custom Poco (lets say, a `User` or an `Account`). In it, is an enum of .. say .. colours. Right now, the values in the JWT are numbers. Yes, it's fine and all ... I'm just OCD. So right now I've enabled the option to pass in a `JsonSerializer` to my `CustomJwtJsonSerializer` (a copy of that linked file, above) which now prints the string version of my claims, out.

I totally get it's just eye candy .. but would you guys be happy to at least review a PR to see if that might be handy to another OCD asshat like me?

(I use jwt.io to decode the jwt to see what contents are in there -> and numbers don't help me see what enums values are, etc).

Thanks kindly 😄

abatishchev:
Hi!
So you're wanting to have an overload added? Or a class?
Can you provide an example?

nbarbettini:
I'm assuming @PureKrome is talking about an overload. By passing a custom `JsonSerializer` the user could customize how things are serialized (by providing custom `JsonConverter`s, and so on).

abatishchev:
I don't mind adding an overload of course (spite adding one to the interface would be a breaking change but who minds?) but not sure how it would look like? Also to have a custom JsonSerializer one just needs to implement the interface differently rather than adding more to the existing implementation. Right?

abatishchev:
Ok, I see now what you're saying. To the addition of existing:
```csharp
        public string Serialize(object obj)
        {
            return JObject.FromObject(obj).ToString(Formatting.None);
        }
```
Add one or more:
```csharp
        public string Serialize(object obj, JsonConverter[] converters)
        {
            return JObject.FromObject(obj).ToString(Formatting.None, converters);
        }

        public string Serialize(object obj, JsonSerializer serializer)
        {
            return JObject.FromObject(obj, serializer).ToString(Formatting.None);
        }

        public string Serialize(object obj, JsonSerializer serializer, JsonConverter[] converters)
        {
            return JObject.FromObject(obj, serializer).ToString(Formatting.None, converters);
        }
```
We need default interface implementation from C# 8.0 here!

PureKrome:
[...]

PR #120, opened 2017-06-23
Adding to JsonNetSerializer ctor accepting JsonSerializer

abatishchev:
Addresses #119.

* Adding to JsonNetSerializer ctor accepting JsonSerializer
* Removing the beta prefix from nuget version
* Replacing xml docs on methods with `<inheritdoc />`
* Soring usings in tests

### DIFF
--- a/src/JWT/Serializers/JsonNetSerializer.cs
+++ b/src/JWT/Serializers/JsonNetSerializer.cs
@@ -30,22 +30,13 @@ namespace JWT.Serializers
             _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
         }
 
-        /// <summary>
-        /// Serialize the given object.
-        /// </summary>
-        /// <param name="obj">The object to serialize.</param>
-        /// <returns></returns>
+        /// <inheritdoc />
         public string Serialize(object obj)
         {
             return JObject.FromObject(obj, _serializer).ToString(Formatting.None);
         }
 
-        /// <summary>
-        /// Deserialize the given string.
-        /// </summary>
-        /// <typeparam name="T">The type to deserialize the string to.</typeparam>
-        /// <param name="json">The JSON to be deserialized.</param>
-        /// <returns></returns>
+        /// <inheritdoc />
         public T Deserialize<T>(string json)
         {
             return JObject.Parse(json).ToObject<T>();
--- a/src/JWT/Serializers/JsonNetSerializer.cs
+++ b/src/JWT/Serializers/JsonNetSerializer.cs
@@ -1,4 +1,5 @@
-﻿using Newtonsoft.Json;
+﻿using System;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace JWT.Serializers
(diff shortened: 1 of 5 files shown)


abatishchev:
@PureKrome will this work for you?

src/JWT/Serializers/JsonNetSerializer.cs
```diff
@@ -8,22 +9,34 @@ namespace JWT.Serializers
...
+        /// <param name="serializer">Internal <see cref="JsonSerializer" /> to use for serialization.</param>
+        public JsonNetSerializer(JsonSerializer serializer)
+        {
+            _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
+        }
+
+        /// <inheritdoc />
+        public string Serialize(object obj)
+        {
+            return JObject.FromObject(obj, _serializer).ToString(Formatting.None);
```
PureKrome: Would the `Formatting.None` override any formatting settings in the `_serializer` ?

src/JWT/Serializers/JsonNetSerializer.cs
```diff
@@ -8,22 +9,34 @@ namespace JWT.Serializers
...
-        /// <typeparam name="T">The type to deserialize the string to.</typeparam>
-        /// <param name="json">The JSON to be deserialized.</param>
-        /// <returns></returns>
+        /// <param name="serializer">Internal <see cref="JsonSerializer" /> to use for serialization.</param>
+        public JsonNetSerializer(JsonSerializer serializer)
+        {
+            _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
+        }
+
+        /// <inheritdoc />
```
PureKrome: TIL: `inheritdoc` 👍

src/JWT/Serializers/JsonNetSerializer.cs
abatishchev: IDK. Can you please try with your settings and see whether it works?

abatishchev: @PureKrome ping

PureKrome: Appols - RL > everything else right now. I'll give it a crack, tonight. watch this space :)

PureKrome: @abatishchev I feel like the formatting is ignored during the `FromObject` stage. And unfortunately, needs to be applied to the `ToString()` stage.

so ...

```
return JObject.FromObject(obj, _serializer).ToString(_serializer.Formatting);
```

BUT! because of me checking out that overload in the `ToString()` .. there's also a converters input param... which means this is also valid.

```
return JObject.FromObject(obj, _serializer).ToString(_serializer.Formatting, _serializer.Converters.ToArray());
```

This is me testing various permutations over the last 30-60 mins. 
I've also ref'd these: 
- https://stackoverflow.com/questions/570689/how-to-serialize-a-jobject-without-the-formatting
- https://stackoverflow.com/questions/20857432/json-net-jobject-fromobject-vs-jsonconvert-deserializeobjectjobjectjsonconver

PureKrome: Oh . that said ...

I couldn't find an online token decoder that RESPECTED the json formatting which was encoded into the JWT. All the online decoders I _think_ formatted the json nicely for me (which makes all of this testing, hardish)

abatishchev: Check it out now

PureKrome: This  LGTM 👍 I'm hoping the additions are actually worth it to other people 😊

abatishchev: One can only hope :-D

abatishchev:
Pushed to NuGet as [3.0.0](https://www.nuget.org/packages/JWT/3.0.0)

abatishchev:
@nbarbettini do you know how it happened that nuget dependency on json.net is now 10.0.0.0? Thought it's 6.4.0.0 (the one that the latest WebApi and MVC have).

PureKrome:
it's been v9 and v10 for a bit....

nbarbettini:
I set it to 10.0.2 in #72, but it looks like it was already there before (in the old project file).

abatishchev:
It seems that v3 always targeted v10 while v2 targets v6

nbarbettini:
I think it was here? https://github.com/jwt-dotnet/jwt/commit/dd3571071c3ab73f57187c09cb7060496574477d#diff-091a88bfcd5c1e8f92f4596998c71e86 (at least 9 -> 10)

abatishchev:
I just might forget. IIRC we were discussing building against the latest version. So it got propagated into the nuspec automatically. But I'd still require not more than 6.0.4 go give people more choices, do not force 10.0 without real need.

abatishchev:
Folks, do you know how binary reference the latest but nuget reference the lowest?

nbarbettini:
> Folks, do you know how binary reference the latest but nuget reference the lowest?

I don't know how to do it in .NET Standard, sorry.

PR #127, opened 2017-07-14
 Modify RS256Algorithm to use RSA from cert directly

### LINKED ISSUE #124: Bug : RS256Algorithm is incorrectly ?
The code should be : (I think...)
```c#
public sealed class RS256JwtAlgorithm : IJwtAlgorithm
    {
        private readonly X509Certificate2 _cert;

        /// <summary>
        /// Creates an instance using the provided certificate.
        /// </summary>
        /// <param name="cert"></param>
        public RS256JwtAlgorithm(X509Certificate2 cert)
        {
            _cert = cert;
        }

        /// <summary>
        /// Signs the provided byte array with the provided key.
        /// </summary>
        /// <param name="key">The key used to sign the data.</param>
        /// <param name="bytesToSign">The data to sign.</param>
        public byte[] Sign(byte[] key, byte[] bytesToSign)
        {
#if NETSTANDARD1_3
            var rsa = (RSACryptoServiceProvider)_cert.GetRSAPrivateKey();
#else
            var rsa = (RSACryptoServiceProvider)_cert.PrivateKey;
#endif
            return rsa.SignData(bytesToSign, "SHA256");
        }

        /// <summary>
        /// The algorithm name.
        /// </summary>
        public string Name => JwtHashAlgorithm.RS256.ToString();
    }
````

No need to create a new RSACryptoServiceProvider?

### PR DESCRIPTION
**Pull request by abatishchev**
Resolves #124, closes #118.

Drawback:

- class `RSA` doesn't have method `SignData` in .NET 3.5, so I upgraded to .NET 4.6.2

### CODE REVIEW

**Comment by abatishchev** [comment]
@nbarbettini can you please check whether I upgraded target framework properly? I'm asking because AppVeyor fails

**Comment by abatishchev** [comment]
@StefH can you please help with test, and what's more important with proper unit testing?

**Comment by StefH** [comment]
Testing can be done like:

1] Generate certificate and private key
```sh
openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout private.key -out certificate_pub.crt
```

2]
Make a X509 certificate based on the two file above (https://github.com/StefH/JwtTest/blob/master/JwtTest/CertificateFromFileProvider.cs)

3]
Generate a token based on RS256 using your old (current) and new (this PR) RS256Algorithm.

4]
Copy paste this token in **Encoded** field on https://jwt.io/

5]
Copy paste your public certificate key (certificate_pub.crt) into that VERIFY SIGNATURE block (first block) only.

6]
When you get a blue box beneath saying **signature verified** then the JWT is valid.

src/JWT/JWT.csproj
```diff
@@ -3,10 +3,9 @@
   <PropertyGroup>
     <TargetFrameworks>net35;netstandard1.3</TargetFrameworks>
```
nbarbettini: This line needs to be `net462;netstandard1.3`

src/JWT/JWT.csproj
```diff
@@ -3,10 +3,9 @@
   <PropertyGroup>
     <TargetFrameworks>net35;netstandard1.3</TargetFrameworks>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
-    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
-    <DefineConstants>NET35</DefineConstants>
-    <FrameworkPathOverride>C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v3.5\Profile\Client</FrameworkPathOverride>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
+    <TargetFramework>net462</TargetFramework>
+    <DefineConstants>NET46</DefineConstants>
```
nbarbettini: Do we have any `#if` directives that look for either `NET35` or `NET46`?

abatishchev: No, only 1 places (this file) that looks for NET_STANDARD. Good point thought!

src/JWT/JWT.csproj
abatishchev: Done, thanks!

abatishchev:
@ubercow sorry, it took some time to come back to the issue you're opened. Can you please take a look on this project? Should be same.

abatishchev:
@StefH right but I meant some input how to update existing/add new unit test(s). Can you please help here?

abatishchev:
AppVeyor can't build anything:
>Could not find Cake.exe at C:\projects\jwt\tools\Cake\Cake.exe
I question its stability :(

nbarbettini:
> AppVeyor can't build anything

That doesn't seem like a typical error - can you try restarting the build?

abatishchev:
It fails with same error for a while now: https://ci.appveyor.com/project/abatishchev/jwt/history

StefH:
@abatishchev 
About unit-tests : you don't really need to test all rsa encryption code. Just use Moq to mock some calls.

src/JWT/JWT.csproj
```diff
@@ -3,10 +3,9 @@
   <PropertyGroup>
     <TargetFrameworks>net35;netstandard1.3</TargetFrameworks>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
-    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
-    <DefineConstants>NET35</DefineConstants>
-    <FrameworkPathOverride>C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v3.5\Profile\Client</FrameworkPathOverride>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
```
abatishchev: Here I want to upgrade to net462. What would be the right way to do that?

abatishchev:
Usually we test end-to-end by having hard-coded JWT and asserting expected outcome

src/JWT/JWT.csproj
StefH: you can just add net462 like:
`<TargetFrameworks>net35;net462;netstandard1.3</TargetFrameworks>`.

But this is not needed because the net35 can be used by all frameworks.

StefH:
If the build works fine, and all code is merged to master branch, I will take a quick look how to do unit test.

src/JWT/JWT.csproj
abatishchev: Currently it's this:

    <TargetFrameworks>net462;netstandard1.3</TargetFrameworks>

Isn't this correct?

abatishchev:
@nbarbettini please take a look to the latest changes

nbarbettini:
I'm not super familiar with the RSACryptoServiceProvider stuff, but it looks reasonable.

Are you intending to raise the minimum framework version to .NET 4.6.2?

abatishchev:
Yeah, I'm not either :) Hope it works, and now works better.

Yes, to normalize the code between .NetFramework and .Net Core. Hope everybody will be fine with that too.

StefH:
If possible, can you lower the dependency for 4.6.2?

Can you lower this to .net 4.0? In that way your librabry can also be used by .net 4 and .net 4.5 libraries/programs.

abatishchev:
Let's continue in #130. Can you please describe your scenario/environment/requirements? The thing from my perspective is that 4.6.2 is the lowest supported as of today version, all lower are out of support. And transition from 4.0 to 4.5. would be long due, and from 4.5 to 4.6.2 usually is seamless. But again I'd love to understand your perspective.

PR #128, opened 2017-07-18
Making JwtEncoder.Validate() public

### LINKED ISSUE #126: Separating Decoding and Validating functionality?
Hello. Thanks for a nice library for JWT.

One thing that I want to request that can you separate the functionality of decoding and validation?

My purpose is that, in my usage, I have a identifier in the payload in which to determine the secret key on the server-side (the recipient of the token), so if I try to decode and validate the token, I have to do the followings.

1. Decode without verification
`var json = decoder.Decode(token, secret: null, verify: false);`
2. Get the identifier from the json payload
`var tokenInfo = JsonConvert.DeseirializeObject<TokenInfo>(json);`
`var identifier = tokenInfo.Identifier;`
3. Get the secret key using the identifier
`var secret = secretsDicionary[identifier];`
4. Decode again, this time, with verification
`decoder.Decode(token, secret: secret, verify: true);`

In order to accomplish the above purpose, can you publicize JwtDecoder.Validate method instead of making it private?
Or maybe creating another interface for validation?

Thank you.

abatishchev:
Resolves #126

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -275,8 +275,8 @@ __pycache__/
 *.pyc
 
 # Cake - Uncomment if you are using it
-# tools/**
-# !tools/packages.config
+tools/**
+!tools/packages.config
 
 # Telerik's JustMock configuration file
 *.jmconfig
--- a/src/JWT/Algorithms/RSAlgorithmFactory.cs
+++ b/src/JWT/Algorithms/RSAlgorithmFactory.cs
@@ -20,9 +20,13 @@ namespace JWT.Algorithms
         /// <inheritdoc />
         public override IJwtAlgorithm Create(JwtHashAlgorithm algorithm)
         {
-            return algorithm == JwtHashAlgorithm.RS256 ?
-                       new RS256Algorithm(_certFactory()) :
-                       base.Create(algorithm);
+            switch (algorithm)
+            {
+                case JwtHashAlgorithm.RS256:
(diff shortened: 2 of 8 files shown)


abatishchev:
@nskhara please validate whether it'd work for you

abatishchev:
My doubt is IValidator becomes very coupled with IDecoder. But I think it's fine. Shall we rework the interface (make public method private and private method public)?

nbarbettini:
I'm not quite following what you mean - but I'm happy to review a PR. 😃

nskhara:
@abatishchev 
Thank you for the change.

One thing, would it be nice if you could add the method into the IJwtDecoder interface also?
(So you can access it through the interface)

`IJwtDecoder decoder = new JwtDecoder(blah, blah, ...);`
`// some other processess...`
`decoder.Validate(blah, blah,...);`

abatishchev:
Yes, that's what we've discussed above and I'll address once AppVeyor is functional again. Later this week for sure since I already started.

abatishchev:
Unfortunately I didn't find this feasible to move this `Validate()` method to `Validator`. I refactored it thought to simplify the signature. Please try yourself or/and let me know if it's fine to merge it as it is for now

abatishchev:
Shall I check this in?

nskhara:
thank you, abatishchev :-)

Issue #138, opened 2017-10-29
missing overloads for decode.


paule96:
Hey hey Developers :)

I have seen a issue in the framework to get the payload of an JWT. To get a payload of an JWT without validation we don't need a key/secret. So in the file [jwt/src/JWT/JwtDecoder.cs ](https://github.com/jwt-dotnet/jwt/blob/master/src/JWT/JwtDecoder.cs) we missed a overload, for the mehtod Decode that only need a token. (for DecodeObject I will create a the same overload)

Can I create a PullRequest for this?

Best regards Paul 👍

paule96:
Here ist the Pullreqeust #139.

abatishchev:
hey Paul, thanks for the changes! let's continue the discussion in the pr itself.

PR #139, opened 2017-10-29
Adding overloads for Decode()

paule96:
Add the missing overloads form #138. 
For better code a add a little Enum for the different parts of an JWT. Maybe we can put this enum to a better place. Lets us discuss about this.

I add some regions in the IJwtDecoder Interface so you can faster find the right methods.

### DIFF
--- a/src/JWT/Jwt.cs
+++ b/src/JWT/Jwt.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Linq;
+
+namespace JWT
+{
+    /// <summary>
+    /// A class that represent a JWT
+    /// </summary>
+    internal class Jwt
+    {
+        /// <summary>
+        /// A array that have length of 3, because the JWT have 3 Parts
+        /// </summary>
+        private readonly string[] _parts;
+
+        /// <summary>
+        /// Creates a new Instance of JWT from a JWT-String
+        /// </summary>
+        /// <param name="token">The JWT as string</param>
+        public Jwt(string token): this(SplitToken(token)) { }
+
+        /// <summary>
+        /// Create a new Instance of a JWT from a part Array of an JWT
+        /// </summary>
+        /// <param name="parts">The parts as Array</param>
+        public Jwt(string[] parts) => _parts = parts;
+
+        /// <summary>
+        /// gets the Header of an JWT
+        /// </summary>
+        public string Header => _parts[(int)JwtParts.Header];
+        /// <summary>
+        /// gets the Payload of an JWT
+        /// </summary>
+        public string Payload => _parts[(int)JwtParts.Payload];
+
+        /// <summary>
+        /// gets the Signature of an JWT
+        /// </summary>
+        public string Signature => _parts[(int)JwtParts.Payload];
+
+        /// <summary>
+        /// gets the Parts of an JWT
+        /// </summary>
+        public string[] Parts => _parts.ToArray();
+
+        /// <summary>
+        /// Split a JWT-String in to its parts
+        /// </summary>
+        private static string[] SplitToken(string token)
+        {
+            var parts = token.Split('.');
+            if (parts.Length != 3)
+            {
+                throw new ArgumentException("Token must consist from 3 delimited by dot parts");
+            }
+            return parts;
+            
+        }
+    }
+
+    /// <summary>
+    /// Helper Enum to get the right part from a JWT-Parts-Array
+    /// </summary>
+    enum JwtParts
+    {
+        Header = 0,
+        Payload = 1,
+        Signature = 2
+    }
+}
--- a/src/JWT/JwtDecoder.cs
+++ b/src/JWT/JwtDecoder.cs
@@ -50,8 +50,18 @@ namespace JWT
         public string Decode(string token, string key, bool verify) => Decode(token, Encoding.UTF8.GetBytes(key), verify);
 
         /// <inheritdoc />
+        /// <exception cref="ArgumentException" />
+        /// <exception cref="ArgumentNullException" />
+        /// <exception cref="ArgumentOutOfRangeException" />
         public string Decode(string token, byte[] key, bool verify)
         {
+            if (String.IsNullOrWhiteSpace(token))
+                throw new ArgumentException(nameof(token));
+            if (key == null)
+                throw new ArgumentNullException(nameof(key));
+            if (key.Length == 0)
+                throw new ArgumentOutOfRangeException(nameof(key));
+
             if (verify)
             {
                 Validate(new JwtParts(token), key);
--- a/src/JWT/JwtDecoder.cs
+++ b/src/JWT/JwtDecoder.cs
@@ -44,56 +44,44 @@ namespace JWT
         }
 
         /// <inheritdoc />
-        public string Decode(string token, string key, bool verify)
+        public string Decode(string token)
         {
-            return Decode(token, Encoding.UTF8.GetBytes(key), verify);
+            var payload = splitToParts(token)[(int)JwtParts.Payload];
+            return Encoding.UTF8.GetString(_urlEncoder.Decode(payload));
         }
+
+        /// <inheritdoc />
+        public string Decode(string token, string key, bool verify) => Decode(token, Encoding.UTF8.GetBytes(key), verify);
 
         /// <inheritdoc />
         public string Decode(string token, byte[] key, bool verify)
         {
-            var parts = token.Split('.');
-            if (parts.Length != 3)
-            {
-                throw new ArgumentException("Token must consist from 3 delimited by dot parts");
-            }
-
-            var payload = parts[1];
-            var payloadJson = Encoding.UTF8.GetString(_urlEncoder.Decode(payload));
-
             if (verify)
             {
-                Validate(parts, key);
+                Validate(splitToParts(token), key);
             }
 
-            return payloadJson;
+            return Decode(token);
         }
 
         /// <inheritdoc />
-        public IDictionary<string, object> DecodeToObject(string token, string key, bool verify)
-        {
-            return DecodeToObject(token, Encoding.UTF8.GetBytes(key), verify);
-        }
+        public IDictionary<string, object> DecodeToObject(string token) => DecodeToObject<Dictionary<string, object>>(token);
 
         /// <inheritdoc />
-        public IDictionary<string, object> DecodeToObject(string token, byte[] key, bool verify)
-        {
-            var payloadJson = Decode(token, key, verify);
-            return _jsonSerializer.Deserialize<Dictionary<string, object>>(payloadJson);
-        }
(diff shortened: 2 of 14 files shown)


src/JWT/JwtDecoder.cs
```diff
@@ -122,5 +110,28 @@ public void Validate(string[] parts, byte[] key)
...
             _jwtValidator.Validate(payloadJson, decodedCrypto, decodedSignature);
         }
+
+        /// <summary>
+        /// Get the Jwt token as string and return a parts array of <see cref="string[]"/> object.
+        /// </summary>
+        /// <param name="token">The JWT.</param>
+        /// <returns>The converted parts array of <see cref="string[]"/></returns>
+        /// <exception cref="ArgumentException">Thrown if the given token have the wrong fromat.</exception>
+        private string[] splitToParts(string token)
```
abatishchev: Please use UpperCase for the name and make it static.

abatishchev: And let's return `IDictionary<JwtParts, stirng>` rather than `string[]` to avoid casting enum to int over and over again.

src/JWT/JwtDecoder.cs
```diff
@@ -102,14 +90,14 @@ public T DecodeToObject<T>(string token, byte[] key, bool verify)
...
+            var crypto = _urlEncoder.Decode(parts[(int)JwtParts.Signature]);
             var decodedCrypto = Convert.ToBase64String(crypto);
 
-            var header = parts[0];
+            var header = parts[(int)JwtParts.Header];
             var headerJson = Encoding.UTF8.GetString(_urlEncoder.Decode(header));
             var headerData = _jsonSerializer.Deserialize<Dictionary<string, object>>(headerJson);
 
-            var payload = parts[1];
+            var payload = parts[(int) JwtParts.Payload];
```
abatishchev: We plan to replace array with dictionary. But please remove space when cast.

src/JWT/JwtDecoder.cs
```diff
@@ -122,5 +110,28 @@ public void Validate(string[] parts, byte[] key)
...
+            var parts = token.Split('.');
+            if (parts.Length != 3)
+            {
+                throw new ArgumentException("Token must consist from 3 delimited by dot parts");
+            }
+            return parts;
+        }
+    }
+
+    enum JwtParts
```
abatishchev: I'm thinking maybe to replace it with a class representing the Jwt? Like:
```csharp
public class JwtPayload
{
    public JwtPayload(string payload) { }

    public JwtPayload(string[] parts) { }

    public string Header { get; } => _parts[0];

    public string Body { get; } => _parts[1];

    public string Signature { get; } => _parts[2];
}
```
So it'd be fine to address by index but only internally.

src/JWT/JwtDecoder.cs
paule96: done

src/JWT/JwtDecoder.cs
paule96: Nice tip. I Resolve this comment.

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            return parts;
+            
+        }
+    }
+
+    /// <summary>
+    /// Helper Enum to get the right part from a JWT-Parts-Array
+    /// </summary>
+    enum JwtParts
```
abatishchev: Do we need the enum now since the addressing by the index is the internal concern of this class in just few places?

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Linq;
+
+namespace JWT
+{
+    /// <summary>
+    /// A class that represent a JWT
+    /// </summary>
+    internal class Jwt
```
abatishchev: `class Jwt` inside `namespace JWT` is quite confusing. Let's have a better name for it. My idea was `JwtPayload` with property `Body`. Another would be `JwtParts` with property `Payload`

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// gets the Parts of an JWT
+        /// </summary>
+        public string[] Parts => _parts.ToArray();
+
+        /// <summary>
+        /// Split a JWT-String in to its parts
+        /// </summary>
+        private static string[] SplitToken(string token)
```
abatishchev: SplitToParts since that's what it is, right?

src/JWT/JwtDecoder.cs
```diff
@@ -102,14 +86,14 @@ public T DecodeToObject<T>(string token, byte[] key, bool verify)
         /// <param name="key">The key that was used to sign the JWT.</param>
         public void Validate(string[] parts, byte[] key)
```
abatishchev:
```csharp
public void Validate(string[] parts, byte[] key) => Validate(new JwtParts(parts), key);

public void Validate(JwtParts parts, byte[] key) // the method itself which accessed the parts by parts.Header,Payload,Signature
```

src/JWT/JwtDecoder.cs
```diff
@@ -122,5 +106,6 @@ public void Validate(string[] parts, byte[] key)
 
             _jwtValidator.Validate(payloadJson, decodedCrypto, decodedSignature);
         }
-    }
+
```
abatishchev: Extra line break :)

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// </summary>
+    internal class Jwt
+    {
+        /// <summary>
+        /// A array that have length of 3, because the JWT have 3 Parts
+        /// </summary>
+        private readonly string[] _parts;
+
+        /// <summary>
+        /// Creates a new Instance of JWT from a JWT-String
```
abatishchev: JWT-String isn't readable imo. Maybe "from the string representation of a JWT"?

src/JWT/Jwt.cs
paule96: I think it makes the code a little better readable. I can removed it but then we have magic numbers in the code. I think this is bad design...

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace JWT
+{
+    /// <summary>
+    /// A class that represent a JWT
+    /// </summary>
+    internal class Jwt
+    {
+        /// <summary>
+        /// A array that have length of 3, because the JWT have 3 Parts
```
abatishchev: "An array that has length of 3 because the JWT has 3 parts"

src/JWT/Jwt.cs
abatishchev: Also I'd make it public since there will be public overloads accepting it.

src/JWT/Jwt.cs
abatishchev: Agreed.

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                throw new ArgumentException("Token must consist from 3 delimited by dot parts");
+            }
+            return parts;
+            
+        }
+    }
+
+    /// <summary>
+    /// Helper Enum to get the right part from a JWT-Parts-Array
```
abatishchev: "from the array representation of JWT parts"?

src/JWT/Jwt.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Linq;
```
abatishchev: Unused `using`?

src/JWT/Jwt.cs
paule96: Something like JwtModel?
Payload isn't correct because the is the word in the RFC document for the Payload part. It's maybe a little bit confusing.

src/JWT/Jwt.cs
paule96: yes.. sorry. can we leave it that way?

src/JWT/Jwt.cs
abatishchev: If the RFC uses the term Payload then we shall stick to it. Model is good when it's something like MVC or MVVP and even there I try not to add obvious suffixes, e.g. Model, Dto, etc. Maybe then just `JwtParts`?

src/JWT/Jwt.cs
paule96: nope. I Return the parts as a new Array with the linq toArray function.

src/JWT/Jwt.cs
abatishchev: Looks from the perspective of internal reader, preferably not that technical. Because this is almost the only documentation we have here. Which one your granny would understand? I'm not a native English speaker so it's little hard me to tell.

Also we can ask @nbarbettini to take a look and suggest, please.

abatishchev:
@paule96 thanks a lot for your contribution! please don't be frustrated (if you're by any chance), I just specialize at work in code style and thoughtful code reviews :)

src/JWT/Jwt.cs
paule96: okay I use now JwtParts. The Enum renamed to JwtPartsPointer

src/JWT/Jwt.cs
paule96: Good idea. I am also not a native english speaker 😃

paule96:
@abatishchev  No it's okay. It's good you have a look of my sometimes bad code :)

src/JWT/JwtDecoder.cs
```diff
@@ -121,6 +105,6 @@ public void Validate(string[] parts, byte[] key)
             var decodedSignature = Convert.ToBase64String(signatureData);
 
             _jwtValidator.Validate(payloadJson, decodedCrypto, decodedSignature);
-        }
-    }
+        }
```
abatishchev: Extra spaces before line breaks. Please click Format Document in VS, press Ctrl+K+D (or K+E depending on the hotkeys layout)

src/JWT/JwtParts.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// An array that has length of 3 because the JWT has 3 parts
+        /// </summary>
+        private readonly string[] _parts;
+
+        /// <summary>
+        /// Creates a new Instance of JWT from the string representation of a JWT
+        /// </summary>
+        /// <param name="token">The JWT as string</param>
+        public JwtParts(string token): this(SplitToken(token)) { }
```
abatishchev: Please format it this way:
```csharp
public JwtParts(string token)
    : this(SplitToken(token))
    {
    }
```
and then re-format the document.

paule96:
Must be done anything else?

abatishchev:
I think we're good. Let's just wait a day @nbarbettini to have a chance to review. Either way I'll merge and release a new version shortly.

abatishchev:
[3.1.0](https://www.nuget.org/packages/JWT/3.1.0) has been pushed to NuGet.

PR #143, opened 2017-11-09
Master based base64url decode padding fix

moljac:
small fix for error in Base64Url Decode with padding (missing 1 for the result of modulo)

Compiled JWT project in VS4Mac and on Windows in VS2017 and with Cake. Green

Could not run Cake on Mac and XUnit tests in VS4Mac. Red


moljac:
My mistake did not read specs properly.
Closing

PR #147, opened 2017-12-21
Adding, renaming, fixing JwtBuilder methods

abatishchev:


src/JWT/Algorithms/HMACSHAAlgorithmFactory.cs
```diff
@@ -25,7 +25,7 @@ public virtual IJwtAlgorithm Create(JwtHashAlgorithm algorithm)
                 case JwtHashAlgorithm.RS256:
                     throw new NotSupportedException($"For algorithm {nameof(JwtHashAlgorithm.RS256)} please create custom factory by implementing {nameof(IAlgorithmFactory)}");
                 default:
-                    throw new InvalidOperationException($"Algorithm {algorithm} is not supported.");
+                    throw new NotSupportedException($"Algorithm {algorithm} is not supported.");
```
nbarbettini: Technically a breaking change. Will this get a version bump?

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -164,101 +187,126 @@ public JwtBuilder SetVerifySignature(bool verify)
...
+                throw new InvalidOperationException("Can't decode a token. Check if you have call all of the followng methods:\r\n" +
+                                                    $"-{nameof(SetSerializer)}\r\n" +
+                                                    $"-{nameof(SetValidator)}\r\n" +
+                                                    $"-{nameof(SetUrlEncoder)}\r\n" +
+                                                    $"If you called {nameof(MustVerifySignature)} you must also call {nameof(SetSecret)}.");
         }
 
         /// <summary>
-        /// Checks whether enought information was supplied to build a new token.
+        /// Checks whether enought dependencies were supplied to build a new token.
```
nbarbettini: nitpick: `enough`

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -164,101 +187,126 @@ public JwtBuilder SetVerifySignature(bool verify)
...
                    _urlEncoder != null &&
                    _jwt.Payload != null &&
-                   _secret != null &&
-                   _secret.Length > 0;
+                   !String.IsNullOrEmpty(_secret);
         }
 
         /// <summary>
-        /// Checks whether enought information was supplied to decode a token.
+        /// Checks whether enought dependencies were supplied to decode a token.
```
nbarbettini: `enough`

src/JWT/Algorithms/HMACSHAAlgorithmFactory.cs
abatishchev: Yes, this and few minor more. So wondering too should we bump to 4.0 or keep 3.2

abatishchev:
@nbarbettini @paule96 another question whether we should rename all builder's method to `With<Bla>()` for consistency across the project and other projects' fluent apis.

nbarbettini:
@abatishchev Like actually using generics, or just `WithIssuer`, `WithAudience`, etc?

abatishchev:
Just `WithIssuer`, `WithAudience`, etc. I think it's more common along similar fluent builders.

nbarbettini:
Yeah, I agree that it's more common. I don't have a strong opinion either way; `WithX` is fine with me.

abatishchev:
Renamed all method to With*

abatishchev:
Pushed to NuGet as [4.0.0-beta1](https://www.nuget.org/packages/JWT/4.0.0-beta1).

PR #151, opened 2017-12-30
Refactoring RS256Algorithm

### LINKED ISSUE #144: Verification with RSA
How to verify JWT signed by RSA key using only the certificate public key?

### LINKED ISSUE #149: JwtBuilder expects secret to be mandatory even for RS256
As per #148.

### LINKED ISSUE #150: RS256Algorithm expects private key to be present, throws NRE if it's not
As per #148.

abatishchev:
* Adding Verify() to RS256Algorithm, resolves #144
* Making secret mandatory only for symmetric algorithms, resolves #149
* Checking certificate private key is not null, resolves #150
* Moving IAlgorithm to Algorithm folder/namespace
* Updating xml docs
* Using predefined constants for target framework in csproj
* Bumping nuget version to beta2


abatishchev:
Published to NuGet as [4.0.0-beta2](https://www.nuget.org/packages/JWT/4.0.0-beta2).

PR #152, opened 2017-12-30
DateTime provider to return DateTimeOffset

### LINKED ISSUE #145: DateTimeOffset should be used rather than DateTime
As @nbarbettini correctly mentioned at https://github.com/jwt-dotnet/jwt/pull/140#issuecomment-351273297, `IDateTimeProvider` should return `DateTimeOffset` rather than `DateTime`. Or both, having the existing method marked as obsolete. Since it'd be a breaking change, it'd require a major version bump.

abatishchev:
Resolves #145.

- DateTime provider to return DateTimeOffset
- Bumping nuget version to beta2

### DIFF
--- a/src/JWT/IDateTimeProvider.cs
+++ b/src/JWT/IDateTimeProvider.cs
@@ -11,6 +11,6 @@ namespace JWT
         /// Get the current DateTime.
         /// </summary>
         /// <returns></returns>
-        DateTime GetNow();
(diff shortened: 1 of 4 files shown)


abatishchev:
@nbarbettini hey Nate, happy new year! can you please to few pull requests I submitted lately? thanks as always!

nbarbettini:
LGTM. Happy new year!

PR #162, opened 2018-03-23
Using single overload of WithSecret() but with params

abatishchev:
* Adding NuGet.config
* Removing, updating packages
* Using single overload of WithSecret() but with params
* Bumping nuget version

### DIFF
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -155,20 +155,7 @@ namespace JWT.Builder
         /// Required to create new token that uses an asymmetric algorithm such as <seealso cref="RS256Algorithm" />.
         /// </remarks>
         /// <returns>Current builder instance</returns>
-        public JwtBuilder WithSecret(string secret)
-        {
-            _secrets = new[] { secret };
-            return this;
-        }
-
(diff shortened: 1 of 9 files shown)


src/JWT/Builder/JwtBuilder.cs
```diff
@@ -155,20 +155,7 @@ public JwtBuilder WithAlgorithm(IJwtAlgorithm algorithm)
...
-
-        /// <summary>
-        /// Sets certificates secrets.
-        /// </summary>
-        /// <remarks>
-        /// Required to create new token that uses an asymmetric algorithm such as <seealso cref="RS256Algorithm" />.
-        /// </remarks>
-        /// <returns>Current builder instance</returns>
-        public JwtBuilder WithSecrets(string[] secrets)
+        public JwtBuilder WithSecret(params string[] secrets)
```
abatishchev: @CodingMate what would you say?

CodingMate: Way better.
How did I miss that, duh :smiley:

abatishchev:
Pushed to NuGet as [5.0.0-beta2](https://www.nuget.org/packages/JWT/5.0.0-beta2).

PR #195, opened 2019-05-05
Change the RS256Algorithm to take in the RSA Service Provider/RSA

### LINKED ISSUE #164: RS256 implemention not having to rely on X509 certificate
Hi,

I am using your JWT token framework to read and interpret AWS Cognito ID tokens.

They use [RS256 signed packets](https://aws.amazon.com/premiumsupport/knowledge-center/decode-verify-cognito-json-token/).

One problem I have is they provide the modulus/exponent directly which allows me to create RSA objects directly, and I don't need the certificate. I wish there was a overload in the factory that allowed you to just provide the RSA provider directly.

Thanks,
Glenn

glennawatson:
This fixes #164

This changes the RS256Algorithm class not to directly take in a Certificate but instead get the RSA/RSACryptoServiceProvider combination.

this is helpful in scenarios such as AWS Cognitio where they provide the modulo and exponent  directly.

### DIFF
--- a/src/JWT/Algorithms/RS256Algorithm.cs
+++ b/src/JWT/Algorithms/RS256Algorithm.cs
@@ -8,14 +8,18 @@ namespace JWT.Algorithms
     /// </summary>
     public sealed class RS256Algorithm : IJwtAlgorithm
     {
-        private readonly X509Certificate2 _cert;
+        private readonly RSACryptoServiceProvider _publicKey;
+        private readonly RSA _privateKey;
 
         /// <summary>
         /// Creates an instance using the provided certificate.
         /// </summary>
-        public RS256Algorithm(X509Certificate2 cert)
+        /// <param name="publicKey">The RSA service provider for verifying the data.</param>
+        /// <param name="privateKey">The RSA key for signing the data.</param>
+        public RS256Algorithm(RSACryptoServiceProvider publicKey, RSA privateKey)
         {
-            _cert = cert;
+            _publicKey = publicKey;
+            _privateKey = privateKey;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/JWT/Algorithms/RS256Algorithm.cs
```diff
@@ -43,28 +48,7 @@ public byte[] Sign(byte[] bytesToSign)
...
-        private static RSACryptoServiceProvider GetPublicKey(X509Certificate2 cert)
-        {
-            AsymmetricAlgorithm alg;
-#if NETSTANDARD1_3
-            alg = cert.GetRSAPublicKey();
-#else
-            alg = cert.PublicKey.Key;
-#endif
-            return (RSACryptoServiceProvider)alg;
+            return _publicKey.VerifyData(bytesToSign, "2.16.840.1.101.3.4.2.1", signature);
```
abatishchev: Add a comment what this "magic number means"

glennawatson: Not sure, it was there in the previous code.

abatishchev:
Thanks so much for contributing this change!
The only thing, beside the comments I left above, please bump the version. At least the minor. Or the major, if it's a breaking change, and it might be.

src/JWT/Algorithms/RS256Algorithm.cs
abatishchev: Before it was API/method calls only. Now it passes this OID. Simple comment what it means, or a url to the spec, would work!

glennawatson: Did a google and the string matched the above. So added the comment.

abatishchev:
Argh! I'm not a pro user of GitHub reviews so messed it up. Can't see half of my own comments.

src/JWT/Algorithms/RS256Algorithm.cs
glennawatson: Definitely was passing in the OID previously btw https://github.com/jwt-dotnet/jwt/pull/195/files

glennawatson:
> Argh! I'm not a pro user of GitHub reviews so messed it up. Can't see half of my own comments.

You can click on the "Show Resolved" button in the right hand side and i'll come up.

src/JWT/Algorithms/RS256Algorithm.cs
abatishchev: Right! Sorry, my bad. Was trying to review from my phone. Took my laptop. Much better now.

glennawatson: BTW I don't think that Verify method is used anywhere. And the OID seems to be a non-RSA algorithm. So might be broken.

abatishchev: `Verify()` was added as a request by a user so I'd keep it.

abatishchev:
If only I could figure out why gated build isn't getting kicked off...

glennawatson:
I am on a fork most likely. Azure DevOps tends to be security concious of external forks to avoid man in the middle attacks.

abatishchev:
Yeap, I allow the build to run on forks. Re-invoking it via the command apparently worked. And it has succeeded.

So what would you say? Good to go?

glennawatson:
Go for it.

abatishchev:
Done via https://abatishchev.visualstudio.com/Jwt.Net/_releaseProgress?_a=release-pipeline-progress&releaseId=11. See https://www.nuget.org/packages/JWT/5.1.0 (should appear soon)

PR #198, opened 2019-05-07
Replacing PowerShell-based build with YAML-based

glennawatson:
A branch to play around with azure.net. Will likely rebase before finishing.


.pipelines/build.yml
```diff
@@ -0,0 +1,35 @@
...
+    solution: $(solution)   
+    msbuildArguments: /restore /t:build /p:CreatePackage=true /p:NoPackageAnalysis=true /p:PackageOutputPath=$(Build.ArtifactStagingDirectory)\artifacts
+    configuration: $(BuildConfiguration)
+    maximumCpuCount: true
+
+- task: DotNetCoreCLI@2
+  displayName: Run Tests
+  inputs:
+    command: test
+    projects: '**/*Tests/*.csproj'
```
abatishchev: Do you know why [the latest gated build](https://abatishchev.visualstudio.com/Jwt.Net/_build/results?buildId=83) finds no test projects?

.pipelines/build.yml
```diff
@@ -0,0 +1,35 @@
...
+    arguments: ' -c $(buildConfiguration) --no-build --no-restore'
+
+- task: PublishBuildArtifacts@1
+  displayName: Publish Build Artifacts
```
abatishchev: It doesn't find any artifacts either by some reason.

abatishchev:
Thanks a lot for making these changes! Can you take a look at few comments I left?

abatishchev:
You think it's better to have 2? Can tests and artifacts be fixed within this one?

glennawatson:
No it's fine. We do it on other projects fine

abatishchev:
Test projects are found now but tests fail

abatishchev:
>##[warning]Invalid results file. Make sure the result format of the file 'd:\a\_temp\VssAdministrator_fv-az674_2019-05-07_18_01_00.trx' matches 'VSTest' test results format.

Hmm...

.pipelines/build.yml
abatishchev: Tests are running now

abatishchev:
Wondering why do we have to use MSBuild over `dotnet build`?

.pipelines/build.yml
```diff
@@ -0,0 +1,39 @@
...
+  buildConfiguration: 'Release'
+  dotNetVersion: '2.2.106'
+
+steps:
+- task: DotNetCoreInstaller@0
+  displayName: Install .NET Core v$(dotNetVersion)
+  inputs:
+      version: $(dotNetVersion)
+
+- task: MSBuild@1
```
abatishchev: Wondering why do we have to use MSBuild over `dotnet build`?

.pipelines/build.yml
abatishchev: NuGet package is now produced and published

.pipelines/build.yml
abatishchev: The build is green now. I think this the last open item before it's good to go in.

glennawatson: You can go back to dotnet build. This was when I was trying to resolve the NuGet issues but I solved that by upgrading the project to using project references instead of packages.config

abatishchev:
I don't know how to deal with this error:

>NU1100: Unable to resolve 'System.Reflection.TypeExtensions (>= 4.5.1)' for '.NETStandard,Version=v1.3'.

Asked on [Stack Overflow](https://stackoverflow.com/questions/56043733/nu1100-unable-to-resolve-system-reflection-typeextensions-4-5-1-for-ne).

.pipelines/build.yml
abatishchev: Doesn't work. Maybe because we have here a mix of .NET Core and .NET Framework projects. I'll figure this out in another PR.

glennawatson:
I got rid of those issue locally by using PackageReference

abatishchev:
The only way I was able to get rid of them on the build server is to roll back this change, unfortunately. And basically 2 parallel pipelines: one for Core and another for Framework.

PR #202, opened 2019-06-13
Add additional framework targets

andrewlock:
Thanks for this library! This PR makes minor tweaks to the targeted frameworks to follow best practice advice 🙂

As you've included a .NET Framework 1.x target, best practice advice is to also include a .NET Standard 2.0 target ([as described in this doc](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting)). This avoids bringing in loads of extra packages on non-NETFX frameworks that support .NET Standard 2.0 (e.g. .NET Core)

Unfortunately, .NET Standard support is all screwed up around `net461`. As you already have a `net46` target, all that's needed to fully fix it is the `net472` target.

From a practical point of view, this shouldn't have any impact on consumers, other than simplifying the package graph on .NET Core!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ Package is avaliable via [NuGet](https://nuget.org/packages/JWT). Or you can dow
 
 ## Supported .NET versions:
 - .NET Framework 4.6.0
+- .NET Framework 4.7.2
 - .NET Standard 1.3
+- .NET Standard 2.0
 
 ## Usage
 ### Creating (encoding) token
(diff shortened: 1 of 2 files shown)


abatishchev:
Thanks a lot for your contribution!

andrewlock:
Great, thanks!

abatishchev:
Published to NuGet as [5.2.0](https://www.nuget.org/packages/JWT/5.2.0)

PR #209, opened 2019-07-02
Fix Parameters checking in the RS256Algorithm's constructor

mirsaeedi:
The messages shown in the exception description are not according to the corresponding variable.

### DIFF
--- a/src/JWT/Algorithms/RS256Algorithm.cs
+++ b/src/JWT/Algorithms/RS256Algorithm.cs
@@ -19,8 +19,8 @@ namespace JWT.Algorithms
         /// <param name="privateKey">The RSA key for signing the data.</param>
         public RS256Algorithm(RSACryptoServiceProvider publicKey, RSA privateKey)
         {
-            _publicKey = publicKey ?? throw new InvalidOperationException("Private key is null");
-            _privateKey = privateKey ?? throw new InvalidOperationException("Public key is null");
+            _publicKey = publicKey ?? throw new InvalidOperationException("Public key is null");
+            _privateKey = privateKey ?? throw new InvalidOperationException("Private key is null");
         }
 
         /// <summary>
@@ -29,7 +29,7 @@ namespace JWT.Algorithms
         /// <param name="publicKey">The RSA service provider for verifying the data.</param>
         public RS256Algorithm(RSACryptoServiceProvider publicKey)
         {
-            _publicKey = publicKey ?? throw new InvalidOperationException("Private key is null");
+            _publicKey = publicKey ?? throw new InvalidOperationException("Public key is null");
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


abatishchev:
Oh! Good catch. Thanks for the fix.

src/JWT/Algorithms/RS256Algorithm.cs
```diff
@@ -19,8 +19,8 @@ public sealed class RS256Algorithm : IJwtAlgorithm
         /// <param name="privateKey">The RSA key for signing the data.</param>
         public RS256Algorithm(RSACryptoServiceProvider publicKey, RSA privateKey)
         {
-            _publicKey = publicKey ?? throw new InvalidOperationException("Private key is null");
-            _privateKey = privateKey ?? throw new InvalidOperationException("Public key is null");
+            _publicKey = publicKey ?? throw new InvalidOperationException("Public key is null");
```
abatishchev: Can you please change `InvalidOperationException` to `ArgumentNullException(nameof(paramName)`

src/JWT/Algorithms/RS256Algorithm.cs
```diff
@@ -91,4 +91,4 @@ private static RSACryptoServiceProvider GetPublicKey(X509Certificate2 cert)
             return (RSACryptoServiceProvider)alg;
         }
     }
-}
\ No newline at end of file
+}
```
abatishchev: Please also bump the minor version in JWT.csproj so I could publish this change to NuGet. Otherwise it'll go live only next time (and who knows when it'll happen?)

abatishchev:
Please address two small comments, and I'll merge right away!

src/JWT/Algorithms/RS256Algorithm.cs
```diff
@@ -19,8 +19,8 @@ public sealed class RS256Algorithm : IJwtAlgorithm
         /// <param name="privateKey">The RSA key for signing the data.</param>
         public RS256Algorithm(RSACryptoServiceProvider publicKey, RSA privateKey)
         {
-            _publicKey = publicKey ?? throw new InvalidOperationException("Private key is null");
-            _privateKey = privateKey ?? throw new InvalidOperationException("Public key is null");
+            _publicKey = publicKey ?? throw new ArgumentNullException("Public key is null");
```
abatishchev: Pass parameter name to first argument, rather than message. You can pass both too, see https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=netframework-4.8#System_ArgumentNullException__ctor_System_String_System_String_

mirsaeedi: Thanks! I fixed it.

src/JWT/Algorithms/RS256Algorithm.cs
mirsaeedi: done!

src/JWT/Algorithms/RS256Algorithm.cs
mirsaeedi: done!

mirsaeedi:
> Please address two small comments, and I'll merge right away!

It's done! Please let me know if there is any other problems.

abatishchev:
Awesome, thanks again for your contribution!

PR #210, opened 2019-08-17
Add method to add several claims

pBouillon:
# The "why"

While using this library in my company, we would have liked to use the builder to add several claims in a more readable way.

We would have liked to be allowed to do something such as:

```csharp
var token = new JwtBuilder()
      .WithSecret(secret)
      .AddClaims({
          "my-claim": "my-value",
          "my-claim-2": "my-value-2"
      })
     .Build();
```

# PR content

I just added a method `AddClaims` allowing the user to pass a IDictionnary of `string: object` in it and add each of its key as a new claim and each of its value as claim value.

### DIFF
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using JWT.Algorithms;
 using JWT.Serializers;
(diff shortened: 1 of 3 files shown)


abatishchev:
Hi! Thanks for the contribution! I'm happy to merge it.
Can you please to add simple test, as well as bump the version in Jwt.csproj?

pBouillon:
Sure thing, there it is !

pBouillon:
Hum, any idea why the restoration failed @abatishchev ?

abatishchev:
You've bumped the wrong thing, Json.Net version instead of Jwt.Net own version :)

abatishchev:
Fixed that.
Waiting for the gated build to succeed. And we're good to go.
Thanks again for the change and quick turnaround too!

abatishchev:
Pushed to NuGet as [5.2.3](https://www.nuget.org/packages/JWT/5.2.3)

PR #211, opened 2019-08-18
Add automatic versioning & sourcelink.

glennawatson:
Added nerdbank git versioning

Added sourcelink support

Use directory.build.props to allow for future packages in sources

Migrate to PackageReference for the .net framework tests package

Use globbing to include the tests


src/JWT/JWT.csproj
```diff
@@ -10,20 +10,6 @@
     <TargetFrameworkIdentifier>.NETStandard</TargetFrameworkIdentifier>
   </PropertyGroup>
 
-  <PropertyGroup>
```
glennawatson: These and any other project under the /src folder will get these properties from the Directory.Build.props file.

glennawatson:
SourceLink will modify the debugging information to include a link to the github commit url of the release. So when the user goes to debug the code if a crash or debugging incident happens it will download the source code for them from github and allow them to debug the file as normal with breakpoints etc.

tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
```diff
@@ -1,7 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\xunit.runner.visualstudio.2.4.1\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.4.1\build\net20\xunit.runner.visualstudio.props')" />
-  <Import Project="..\..\packages\xunit.core.2.4.1\build\xunit.core.props" Condition="Exists('..\..\packages\xunit.core.2.4.1\build\xunit.core.props')" />
```
abatishchev: Can you please submit this change as a separate PR? I'd like to check it on immediately while other may require additional discussion and consideration?

version.json
```diff
@@ -0,0 +1,17 @@
+{
+  "version": "5.3",
+  "publicReleaseRefSpec": [
+    "^refs/heads/master$", // we release out of master
+    "^refs/heads/develop$", // we release out of develop
```
abatishchev: Please drop last two as this repo has never been released from branches other master

src/Directory.Build.props
```diff
@@ -0,0 +1,36 @@
...
+    <EnableSourceControlManagerQueries>$(EnableSourceLink)</EnableSourceControlManagerQueries>    
+    <!-- Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl> 
+    <!-- Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Include PDB in the built .nupkg -->
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(SourceLinkEnabled)' != 'false'">
```
abatishchev: Should we enable SourceLink as a separate PR as well?

.pipelines/build.yml
```diff
@@ -29,6 +29,24 @@ steps:
...
+    command: custom
+    custom: tool
+    arguments: install --tool-path . nbgv
+  displayName: Install NBGV tool
+  env:
+    COREHOST_TRACE: 0
+    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
+    DOTNET_CLI_TELEMETRY_OPTOUT: 1    
+
+- script: nbgv cloud
```
abatishchev: That's what I'd like to discuss further: combining build and release. That's what I'd rather do not do. Until there are good reasons. I'd like to learn more.

glennawatson: This is effectively just communicating any version information set in the project to azure DevOps. So if building locally this isn't needed but the version information is still injected. This is the missing step you had issues with where it was essentially just the increment of build numbers instead of version number.

src/Directory.Build.props
glennawatson: Probably. I will likely do four pr then. Introduce directory.build.props, introduce source link, introduce nbgv, change to package reference.

.pipelines/build.yml
abatishchev: What I'd like to have is YAML-based release definition but I don't see it's easily available/doable. They introduced something called multi-stage YAML-based build definitions instead, which suppose to replace release. Are you familiar with this area?

src/JWT/JWT.csproj
abatishchev: But we have just 1 project under /src, should we move anything to a common fille? Won't it hurts readability/maintainability?

src/Directory.Build.props
abatishchev: Please, much much appreciate your participation and contributions!

src/JWT/JWT.csproj
glennawatson: I did it mostly to increase maintainability since in my experience these settings get out of sync fast when a new project gets introduced

abatishchev: I totally agree for general case, and that's what I do at my work projects too. But this project haven't seen more than packageable project since its inception and very unlikely will see it in the future. So I'd like to keep it simple and park for now. As soon if we'll decide to have more than one project - we'll have a common version file.

glennawatson:
I'm going to return to the pr feedback on Thursday for you.

glennawatson:
Will do a separate PR for nerdbank with the feedback provided. This has become a PR just about sourcelink and the supporting csproj tags.

glennawatson:
The source link one. Yep ready for you to merge. Thanks. 

I will play around with the azure DevOps feature you want to use for the Nerdbank versioning but will be a week before I can get back to it.

PR #212, opened 2019-08-18
Cleanup unit tests

pBouillon:
# Goal

## How
I refactored most of the tests to normalize their redaction. Using the [standard AAA testing](https://www.c-sharpcorner.com/UploadFile/dacca2/fundamental-of-unit-testing-understand-aaa-in-unit-testing/) and two tools to simplify them as much as possible.

### Fluent Assertions
I used [FluentAssertions](https://fluentassertions.com/), most of the tests are now way more readables and may surely allow newcomers to understand better and faster how to understand the underlying logic.

For example:

```csharp
// From this test case:
Assert.True(token.Length > 0 && token.Split('.').Length == 3);

// To this one:
tokenBuilt.Should()
     .NotBeEmpty("because the token should contains some data");

tokenBuilt.Split('.').Should()
     .HaveCount(3, "because the built token should have the three standard parts");
``` 

### AutoFixture

[https://github.com/AutoFixture/AutoFixture](Autofixture) is a tool that allow us to specify a data type that we are going to use, without having to handle its initialization and writting magic data.  It also bring more randomness in the unit tests and allow us to have a behaviour less predictable.

With this package, we can be focused only on the data we are working with and that matter for the test case.

For example:

```csharp
// From this test case:
    .WithSecret("fjhsdghflghlk")

// To this one:
var secret = _fixture.Create<string>();
// ...
    .WithSecret(secret)
```

### DIFF
--- a/tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
+++ b/tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>JWT.Tests.NETFramework</RootNamespace>
     <AssemblyName>JWT.Tests.NETFramework</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">15.0</VisualStudioVersion>
--- a/tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
+++ b/tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>JWT.Tests.NETFramework</RootNamespace>
     <AssemblyName>JWT.Tests.NETFramework</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.7.3</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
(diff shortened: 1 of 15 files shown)


tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>JWT.Tests.NETFramework</RootNamespace>
     <AssemblyName>JWT.Tests.NETFramework</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
```
abatishchev: Can you please explain this change?

abatishchev:
Thank you so much! I'm always using FA in tests myself and often use AF too. Tests in this project needed more love for a long time.

The only change I'm not sure about is the downgrade from net462 to net61.

tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj
pBouillon: Oh yeah, you're right, Visual Studio 2019 upgraded my fork of the project, I'm going to remove it

abatishchev: Cool! Can you try net473, would it work on the build server? I always love to be on the version, especially in tests.

pBouillon: sure thing ! I'm doing it right now !

pBouillon: Should be done !

pBouillon:
What is this failure @abatishchev ? 🤔

abatishchev:
I think the latest is net472, not 473. Can you please try that one? It should succeed then and I'll merge right away

pBouillon:
Done !

abatishchev:
Yay, thank you again!

pBouillon:
You're welcome ! Hope I can help in a near future !

abatishchev:
If you'd like to contribute more, please help to convert the project with the .NET Framework version of tests to PackageReference/SDK-based, one of the item @glennawatson has planned to do in #211.

Another option is to get rid off then altogether as I'm not quite sure they bring much value.

What both of you think about this?

pBouillon:
I'm not very familiar with this but I would be glad to help !  
Could you open an issue with more details so I can search a little how to do it ?

glennawatson:
The work has already been done just needs to separated into a separate pr

glennawatson:
https://github.com/jwt-dotnet/jwt/blob/84b4108d673b93eaee1aa5220fab14e2ff469d9f/tests/JWT.Tests.NETFramework/JWT.Tests.NETFramework.csproj is the file. Also delete the packages.config file

pBouillon:
I'll give it a look whenever I can

Issue #220, opened 2019-09-26
Extend RSAlgorithmFactory to support all ctors of RS256Algorithm


abatishchev:
Per the discussion in #173#issuecomment-535515008

flerka:
added draft pr for this

PR #221, opened 2019-09-26
Extend RSAlgorithmFactory to support all ctors of RS256Algorithm

flerka:
This is my draft implantation for https://github.com/jwt-dotnet/jwt/issues/220

### DIFF
--- a/src/JWT/Algorithms/RSAlgorithmFactory.cs
+++ b/src/JWT/Algorithms/RSAlgorithmFactory.cs
@@ -14,14 +14,11 @@ namespace JWT.Algorithms
         /// </summary>
         /// <param name="certFactory">Func that returns <see cref="X509Certificate2" /> which will be used to instantiate <see cref="RS256Algorithm" /></param>
         public RSAlgorithmFactory(Func<X509Certificate2> certFactory) =>
-            _algFactory = () =>new RS256Algorithm(certFactory());
+            _algFactory = () => new RS256Algorithm(certFactory());
 
         /// <summary>
         /// Creates an instance of <see cref="RSAlgorithmFactory"/> using the provided public key only.
         /// </summary>
-        /// <remarks>
-        /// An instance of <see cref="RSAlgorithmFactory" /> created using this constructor can only be used for verifying the data, not for signing it.
-        /// </remarks>
         /// <param name="publicKey">The public key for verifying the data.</param>
         public RSAlgorithmFactory(RSA publicKey) =>
             _algFactory = () => new RS256Algorithm(publicKey);
--- a/src/JWT/Algorithms/RSAlgorithmFactory.cs
+++ b/src/JWT/Algorithms/RSAlgorithmFactory.cs
@@ -26,9 +26,6 @@ namespace JWT.Algorithms
         /// <summary>
         /// Creates an instance of <see cref="RSAlgorithmFactory"/> using the provided pair of public and private keys.
         /// </summary>
-        /// <remarks>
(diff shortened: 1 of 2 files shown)


abatishchev:
Hi, thanks for the contribution!
Can you explain the idea behind heaving `Func` over instantiating an algorithm directly?

flerka:
This was done to avoid confusing if inside switch statement. All Func can be rewritten with:

```csharp
            switch (algorithm)
            {
                case JwtHashAlgorithm.RS256:
                    {
                        if (_publicKey != null)
                        {
                            return new RS256Algorithm(publicKey);
                        }

                        var certificate = _certFactory();
#if NETSTANDARD1_3
            return new RS256Algorithm((RSACryptoServiceProvider)certificate.GetRSAPublicKey(), certificate.GetRSAPrivateKey());
#else
                        return new RS256Algorithm((RSACryptoServiceProvider)certificate.PublicKey.Key, (RSA)certificate.PrivateKey);
#endif

                    }
                default:
                    throw new NotSupportedException($"For algorithm {Enum.GetName(typeof(JwtHashAlgorithm), algorithm)} please use the appropriate factory by implementing {nameof(IAlgorithmFactory)}");
            }
```
If this is better from your point of view, I can rollback previous changes and commit this. What should I do @abatishchev ?

abatishchev:
I was thinking more about this factory and realized we might not need it at all. Since it supports a single algorithm anyway. So it just mimics the algorithm's own constrictors. Does it add any value in top of it?
The initial ideas was to support multiple algorithms, help to instantiate one or another. But that has never happened.
So I'd remove it together. What do you think?

abatishchev:
I'd remove `JwtHashAlgorithm` as well. Wondering what's the percentage of the users of this library uses symmetric algorithms still. Hope that not much.

abatishchev:
Actually I got your idea and I like it. It's better than mine in #223. Can you please rebase on the latest master?

src/JWT/Algorithms/RSAlgorithmFactory.cs
```diff
@@ -8,31 +8,48 @@ namespace JWT.Algorithms
...
+        {
             _certFactory = certFactory;
+            _createAlgorithmMethod = CreateWithCertificate;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RSAlgorithmFactory"/> class
+        /// </summary>
+        /// <param name="publicKey">The RSA service provider which will be used to instantiate <see cref="RS256Algorithm" /></param>
+        public RSAlgorithmFactory(RSACryptoServiceProvider publicKey) =>
```
abatishchev: Let's accept just `RSA` here

src/JWT/Algorithms/RSAlgorithmFactory.cs
```diff
@@ -8,31 +8,48 @@ namespace JWT.Algorithms
...
     {
         private readonly Func<X509Certificate2> _certFactory;
+        private readonly Func<RS256Algorithm> _createAlgorithmMethod;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RSAlgorithmFactory"/> class
         /// </summary>
         /// <param name="certFactory">Func that returns <see cref="X509Certificate2" /> which will be used to instantiate <see cref="RS256Algorithm" /></param>
-        public RSAlgorithmFactory(Func<X509Certificate2> certFactory) =>
+        public RSAlgorithmFactory(Func<X509Certificate2> certFactory)
```
abatishchev: Will also need a ctor that accepts both `RSA publicKey` and `RSA privateKey`

abatishchev:
@flerka thanks for your contribution, much appreciated.

Issue #239, opened 2019-12-30
Stuck trying to create an RS256 bearer token : unable to create an RS256Algorithm


PureKrome:
👋 G'Day!

Ok, I'm trying to create an RS256 token. When I do this, the `var algorithm = new RS256Algorithm(certificate);` throw an exception:

```
   System.ArgumentNullException : Value cannot be null.
   Parameter name: privateKey
```

This is because (I think) the [certificate doesn't have a private key](https://github.com/jwt-dotnet/jwt/blob/master/src/JWT/Algorithms/RS256Algorithm.cs#L44)?

This is the code I'm using to generate a my "encode" logic:

```
var certificateText = File.ReadAllText("Data\\PublicCertificateFromAuth0.txt");
var publicCertificate = Encoding.ASCII.GetBytes(certificateText);
var certificate = new X509Certificate2(publicCertificate);
 
var algorithm = new RS256Algorithm(certificate);
<the rest is snipped>
```

and the certificate text is taken from Auth0 .. here ..

![image](https://user-images.githubusercontent.com/899878/71576277-3da1b600-2b44-11ea-89c4-e732ff1e1dc0.png)

So then I thought "Ok, just add the client secret, for this Auth0 account. Surely that's the password for this certificate, right?"

```
var password = "auth0 client secret";
var certificate = new X509Certificate2(publicCertificate, password);
```

(and this is where I got the "client secret" from)...

![image](https://user-images.githubusercontent.com/899878/71576351-7b9eda00-2b44-11ea-9d21-7b119dc7487b.png)

But when I try that (above), the instance of the certificate has the private key as `null`, still.

![image](https://user-images.githubusercontent.com/899878/71576416-be60b200-2b44-11ea-930f-774e7618487f.png)

So .. I'm not really sure what's going on or what to do here.

SIDE NOTE: I'll try and submit some PR for some notes/readme update because it looks like RS256 is becoming the norm for these things these days. That said, it's mostly for decoding, not encoding.

abatishchev:
If you don't have private key then construct RSA and pass it to the algorithm directly, don't attempt to construct a certificate first and then pass it to the algorithm.

Yes, I agree, asymmetric encryption is widely used while almost no one these days needs hashing.

abatishchev:
See [RS256Algorithm.cs](https://github.com/jwt-dotnet/jwt/blob/master/src/JWT/Algorithms/RS256Algorithm.cs) for available ctors.

[...]

PR #240, opened 2019-12-31
RS256 ctor can accept cert with no private key

PureKrome:
- I'm not sure how to create a certificate _without_ a public key. (I would like to test this scenario, if it's doable).
- Included some tests for
  - Certificate has public key only.
  - RS256Algorithm can be created with only a public key (existing code, just adding a simple test).

### DIFF
--- a/src/JWT/Algorithms/RS256Algorithm.cs
+++ b/src/JWT/Algorithms/RS256Algorithm.cs
@@ -42,14 +42,7 @@ namespace JWT.Algorithms
         /// <param name="cert">The certificate having a public key and an optional private key.</param>
         public RS256Algorithm(X509Certificate2 cert)
         {
-            if (cert is null)
-            {
-                throw new ArgumentNullException(nameof(cert));
-            }
-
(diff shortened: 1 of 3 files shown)


abatishchev:
Awesome! Please bump the nuget package version inside JWT.csproj. And it's good to go.

PureKrome:
To what version? (i'm not sure what your versioning / semver rules are etc).

to: `6.1.0-beta1` or `6.1.0`

abatishchev:
Since beta was never published before and based in prior versioning scheme, I'd call it 6.0.0-beta2.

PureKrome:
Currently, it's `6.0.0-beta2` so i'll make it `beta3`

PureKrome:
Out of interest - is there a way to
- create a certificate without a public key? (i'm guessing, nope?)
- test with a certificate that has public/private key. I'm just using the `TestData.PublicKey1 or 2` to test ... is there a way to add in something with a public and private key, to test that?

abatishchev:
Helps with #239

PR #252, opened 2020-02-08
Make AddCalims accept an IReadOnlyDictionary

IvaYan:
The `AddClaims()` method does not modify the payload in any way. We can use `IReadOnlyDictionary` instead of ordinary `IDictionary`.

### DIFF
--- a/src/JWT/Builder/JwtBuilderExtensions.cs
+++ b/src/JWT/Builder/JwtBuilderExtensions.cs
@@ -28,7 +28,7 @@ namespace JWT.Builder
         /// <summary>
         /// Adds several claims to the JWT
         /// </summary>
-        public static JwtBuilder AddClaims(this JwtBuilder builder, IDictionary<string, object> claims) =>
+        public static JwtBuilder AddClaims(this JwtBuilder builder, IReadOnlyDictionary<string, object> claims) =>
             claims.Aggregate(builder, (b, p) => b.AddClaim(p.Key, p.Value));
 
         public static JwtBuilder ExpirationTime(this JwtBuilder builder, DateTime time) =>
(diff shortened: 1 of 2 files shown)


abatishchev:
Hi! I'm happy to make this change.
But first can you please explain what's the issue right now: does it work or not; if yes then does it misbehave or what?

IvaYan:
Hi!
Everything works fine and behaves as expected, but to me it seems odd that we pass a mutable dictionary in order to simply read its content.

For example, we may have an `AccessToken` class like this:

```C#
class AccessToken {
    public IReadOnlyDictionary<string, object> Payload {get;}
}
```

This class manages its state internally, so it exposes the payload as a read-only dictionary so that others can't modify it. With the current implementation we can't use it like this
```C#
var encodedToken = new JwtBuilder()
                .WithSecret(Secret)
                .AddClaims(Payload)
                .Build();
```
because `AddClaims()` does not accept `IReadOnlyDictionary`.

abatishchev:
Sure, it makes sense to pass something more lightweight. Would `IEnumerable<KeyValuePair<K,V>>` work/work better here?

IvaYan:
Yeah, I agree, `IEnumerable` fits even better here. Shall I start another pull request?

abatishchev:
I'd just edit this PR. And please bump the version in JWT.csproj to 6.0.1 so I'll publish it right away.

IvaYan:
Done.

abatishchev:
Sorry, please change only Version, other should stay 6.0.0.0 (simplifies many upgrade and debug scenarios).

abatishchev:
Yay, thanks for the contribution! I'll publish it shortly.

Issue #14, opened 2015-03-10
Assembly does not have a strong name


MikeChristensen:
This assembly is not signed with a key, which makes it impossible to use with assemblies that must be GAC'ed.

darrelmiller:
Microsoft are no longer Strong Naming assemblies in ASP.NET 5.  Their guidance has changed regarding strong naming.
http://stackoverflow.com/questions/28352756/microsoft-aspnet-http-does-not-have-a-strong-name

MikeChristensen:
Oh wow, that I didn't know!  I wonder if that means that you'll be able to
GAC things that aren't signed in v.Next.  Anyway, we were able to work
around the limitation.  We now ilmerge the binaries we need into one
assembly, and then sign that.  Thanks for getting back to me!

darrelmiller:
For ASP.NET Core CLR there is no more GAC, for the full desktop CLR there is a GAC and I recall the signing rules are different, but I don't know the details off hand.

mikelehen:
Thanks for chiming in, @darrelmiller.  I'm closing this out for now.  If anybody's blocked by this, feel free to holler and we can revisit.

balajikandasamy:
I am using an assembly https://github.com/salesforce-marketingcloud/FuelSDK-CSharp that's dependent on JWT. If the assembly is not signed I am not able to deploy the dll to GAC. The FuelSDK is also not signed and the team has agreed to release it out next week. The StackOverflow post mentioned above has been updated and the assemblies are signed in ASP.NET 5 now. Ref: https://github.com/salesforce-marketingcloud/FuelSDK-CSharp/issues/70

abatishchev:
hi, was Jwt.Net ever signed? No, to my knowledge. And I don't see (FuelSDK-CSharpSDK)[https://www.nuget.org/packages/FuelSDK-CSharpSDK] depends on [Jwt.Net](https://www.nuget.org/packages/JWT/).

balajikandasamy:
Please check this one is the nugget package I'm taking about
https://www.nuget.org/packages/SFMC.FuelSDK/

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

abatishchev:
Please submit a PR to sign it and bump the version to 6.0.0-beta1.

abatishchev:
@balajikandasamy ping. Please contribute if you'd like to see this happen.

balajikandasamy:
Sure I will do that this week.
Thanks,
[...]

PR #253, opened 2020-02-11
Provide a strong name for all assemblies

FishDawg:
Provide a strong name for all assemblies. This allows them to be referenced by other assemblies that have a strong name. This is a good practice. Fixes issue #14.


src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
```diff
@@ -15,6 +15,8 @@
     <AssemblyVersion>6.0.0.0</AssemblyVersion>
```
abatishchev: Please bump Version for all 3 packages but don't bump the other 2 attributes.

src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
```diff
@@ -15,6 +15,8 @@
...
     <RootNamespace>JWT</RootNamespace>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <SignAssembly>true</SignAssembly>
```
abatishchev: Can this be centralized using Directory.Build.props? See https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets
You can drop it /src or the root (see my another comment)

tests/JWT.Extensions.AspNetCore.Tests/JWT.Extensions.AspNetCore.Tests.csproj
```diff
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <SignAssembly>true</SignAssembly>
```
abatishchev: Do we really need to sign tests? I don't think we do. So having the file on /src would skip signing tests.

FishDawg: Yeah, that's fair. I don't personally have a need to reference the test assemblies in another project. I'd probably just give them a strong name anyway because it doesn't hurt anything and can only potentially help. But I'm fine either way. What do you think?

abatishchev: I don't have a preference really either. But I see the props file already exists under /src and having a hierarchy of them is little more complex than having a single one. So that's the only reason I'd go with signing only under /src.

FishDawg: Ah, I didn't see the file there. Yes, that makes sense to me too.

src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
FishDawg: Should I make it 6.1.0.0? Not sure if you'd consider this a breaking change. It could break someone who is manually referencing/loading the assembly or otherwise relying on the assembly's full name.

tests/JWT.Extensions.AspNetCore.Tests/JWT.Extensions.AspNetCore.Tests.csproj
FishDawg: Actually, I found a problem with that. The InternalsVisibleTo attribute is being used and it requires the specified assembly to have a strong name (when its own assembly has a strong name). So, I think signing the test assemblies is best.

FishDawg: It seems it doesn't like having Directory.Build.props at multiple levels. So, I'll add the property in both the /src and /tests directories.

abatishchev: Here's how you can make it respecting the hierarchy of such files: add this line to the top of the nested file:
```xml
  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)..\'))" />```

File: src/Directory.Build.props [hunk 4]
```diff
@@ -11,4 +11,8 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
+  <PropertyGroup>
```
abatishchev: Or another idea - we can just can move this file to the root and that's it.

FishDawg: Yeah, I think that's good. Take a look when you get a chance. And is 6.1.0.0 good for the version number?

abatishchev: Yes, I think we need to bump at least the major version for the main package to 6.1.0 and probably keep the other two 6.0.0. And bump the build version for 2 extensions packages.

FishDawg: The other two are currently version 6.0.0-alpha1. Just leave that? And I don't know what you mean by the build version.

abatishchev: Hmm please bump those two to `6.0.0-alpha2`

src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
abatishchev: Let's keep 6.0.0.0, I think that will be fine.

abatishchev:
Awesome, thanks! Let's wait for the gated build to finish and ship it.

PR #258, opened 2020-03-23
Fixed bug in HasOnlyOneSecret(), updated tests

abatishchev:
Fixes #256 

* Fixed bug in HasOnlyOneSecret()
* Marked JwtBuilder.Build() with error: true
* Forked .NET Core tests into 2.2 and to 3.1
* Bumped version to 6.1.1

PR #259, opened 2020-03-24
Made secret optional in Decode() to support asymmetric algorithms

abatishchev:
Continuation of #258, cc @stefansolid.

* Made secret optional in Decode() to support asymmetric algorithms
* Bumped version to 6.1.2

### DIFF
--- a/src/JWT/JWT.csproj
+++ b/src/JWT/JWT.csproj
@@ -15,7 +15,7 @@
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>6.1.1</Version>
+    <Version>6.1.2</Version>
     <FileVersion>6.0.0.0</FileVersion>
     <AssemblyVersion>6.0.0.0</AssemblyVersion>
     <RootNamespace>JWT</RootNamespace>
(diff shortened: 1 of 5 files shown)


abatishchev:
@stefensolid please take a look and try [this build](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=407&view=artifacts&type=publishedArtifacts)'s drop, see if resolves the rest of your issues.

stefansolid:
Now the builder can run MustVerifySignature if the cert is the private cert. 

However, when I on the client uses the MustVerifySignature and only providing the public cert I get the same exception (Exception thrown: 'System.InvalidOperationException' in JWT.dll)

So almost there :)

abatishchev:
The library verified the signature by creating it and comparing with the original. And for that you need the private key. Because if you could do that with having only the public key then you could forge the signature and defeat the whole security mechanism of JWT. So the need to have the private key sounds like by design to me.

stefansolid:
Seems logically, I admit this is not my strongest of skills. I just took my JWT created and posted it on jwt.io where there was a box to enter "Public Key or Certificate. Enter it in plain text only if you want to verify a token". So it was from that I got idea that token could be verified with public key.

abatishchev:
Hmm, you're right. I see you [can do it](https://stackoverflow.com/questions/34403823/verifying-jwt-signed-with-the-rs256-algorithm-using-public-key-in-c-sharp) in C# too. But differently than the library is doing it today.
Let me finish this PR and I'll be working to make it possible.

stefansolid:
Seems to work just perfect now! My client loads the public cert and passes validation without exception. Server-side uses the private cert for creation and validation of tokens.

abatishchev:
Perfect! Thanks for the confirmation. I'm merging it now.

PR #261, opened 2020-03-25
Allowed to verify signature with only public key

abatishchev:
* Allowed to verify signature with only public key
* Bumped version to 6.1.3
* Updated test


abatishchev:
@stefansolid please try [this build](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=418&view=artifacts&type=publishedArtifacts)

abatishchev:
@stefansolid just to confirm, you were referring to the build from this PR, the latest 6.1.3.

stefansolid:
Correct, downloaded binaries this morning, 6.1.3 is DLL-version number, just checked.

PR #264, opened 2020-04-02
Support building decoder without algorithm factory

### LINKED ISSUE #263: JwtDecoder throws error if no Algorithm/Factory specified even though it's not needed
I need to decode a token in order to get a keyid, so that I can go download the public key, so that I can specify an algorithm.  Bit of a chicken/egg problem.

Looking into the JwtDecoder I can see that the Algorithm is only actually needed when verifying/validating, so the builder needn't throw an error where the Alg/Factory is not specified, particularly if `DoNotVerifySignature()` has been called.

To make this work I have had to use a Mock `IJwtAlgorithm` :/

abatishchev:
Resolves #263.

* Added ctor to JwtDecoder without validator and algorithm
* Updated builder
* Updated tests
* Bumped version to 6.1.4


abatishchev:
@richardhauer can you please the package from [this build](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=436&view=results) before I'll publish it?

Can you please also contribute back your `NoOpAlgorithm`, just in case someone else might need it in the future?

abatishchev:
@richardhauer please try [6.1.4](https://www.nuget.org/packages/JWT/6.1.4)

PR #269, opened 2020-04-05
Added JwtDecoderContext, moved extensions and exceptions, renamed JwtHashAlgorithm

### LINKED ISSUE #268: Need an AlgorithmFactory delegate that receives some header/body claims as arguments
I need a delegate that can receive unverified claims during the call to the algorithm factory. Something like .WithAlgorithmFactory() but an overload that includes unverified claims, specifically the `iss` from the body, and the `keyId` and the `typ` from the header.

In my case I need to resolve JWTs generated by more than one system, so I don't know the RSAParameters in advance - I need the `iss` so I can get the retieve the WellKnownConfiguration, so I can get the JwkUrl, which I then cache. The Jwk has RSAParameters (e/n) that are matched to the `keyId` so I can dynamically validate JWTs as long as I know where to go to get the WellKnownConfiguration, based on the Issuer.

For this I have an IRsaResolver::GetRsa( string issuer, string keyId ) but I can't wire that into the JwtBuilder as far as I can see. My implementation then caches the RSAParameters against the $"{Issuer}/{KeyId}" for 24hrs for performance.  

For me these are always RSA so the `typ` claim isn't needed, but a more generic implementation might have to account for non-RSA signatures and the JWT does advertise the correct method to use, so it might as well be included in the argument list.

Maybe an idea for v8

abatishchev:
Resolves #268 

* Added JwtDecoderContext
* Changed DecodeHeader()
* Converted some methods of IAlgorithmFactory to extension methods
* Moved extension methods to file with interface
* Moved exceptions to folder Exceptions
* Renamed JwtHashAlgorithm to JwtAlgorithmName, moved to folder Algorithms
* Bumped version to 7.0.0-beta3


abatishchev:
@richardhauer please try this out and let me know if it works for you.

PR #271, opened 2020-04-22
Fixed sample code in readme to use Encode()

alongthepike:
Sample code for Encoding using the Fluent API in the ReadMe was still using Build() which is obsolete instead of Encode().  Fixed.

Thanks for the great and easy to use library!


abatishchev:
Cool! Thanks for the contribution and for using the library :)

PR #283, opened 2020-05-15
Replace Newtonsoft.Json with System.Text.Json

### LINKED ISSUE #218: Replace Newtonsoft.Json with System.Text.Json
Consider switching to system.text.json for the serilialiser and removing newtonsoft dependency.

luposky:
Now (only in netstandard20) you can use SystemTextJsonSerializer as a serializer.
Some limitations apply, resolves #218.

### DIFF
--- a/src/JWT/Builder/JwtHeader.cs
+++ b/src/JWT/Builder/JwtHeader.cs
@@ -1,4 +1,7 @@
 ﻿using Newtonsoft.Json;
+#if SYSTEMTEXTJSON
+using System.Text.Json.Serialization;
+#endif
 
 namespace JWT.Builder
 {
@@ -8,24 +11,45 @@ namespace JWT.Builder
     public class JwtHeader
     {
         [JsonProperty("typ")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("typ")]
+#endif
         public string Type { get; set; }
 
         [JsonProperty("cty")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("cty")]
+#endif
         public string ContentType { get; set; }
 
         [JsonProperty("alg")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("alg")]
+#endif
         public string Algorithm { get; set; }
 
         [JsonProperty("kid")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("kid")]
+#endif
         public string KeyId { get; set; }
 
         [JsonProperty("x5u")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("x5u")]
+#endif
         public string X5u { get; set; }
 
         [JsonProperty("x5c")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("x5c")]
+#endif
         public string X5c { get; set; }
 
         [JsonProperty("x5t")]
+#if SYSTEMTEXTJSON
+        [JsonPropertyName("x5t")]
+#endif
         public string X5t { get; set; }
     }
 }
--- a/src/JWT/JWT.csproj
+++ b/src/JWT/JWT.csproj
@@ -36,8 +36,16 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <DefineConstants>SYSTEMTEXTJSON;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
+  </ItemGroup>
+
   <ItemGroup>
-   <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3' OR '$(TargetFramework)' == 'netstandard2.0'">
@@ -49,4 +57,5 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
+
 </Project>
--- a/src/JWT/JWT.csproj
+++ b/src/JWT/JWT.csproj
@@ -18,9 +18,9 @@
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>7.2.0</Version>
-    <FileVersion>7.0.0.0</FileVersion>
-    <AssemblyVersion>7.0.0.0</AssemblyVersion>
+    <Version>8.0.0-beta1</Version>
+    <FileVersion>8.0.0.0</FileVersion>
+    <AssemblyVersion>8.0.0.0</AssemblyVersion>
     <RootNamespace>JWT</RootNamespace>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
   </PropertyGroup>
--- a/src/JWT/Serializers/SystemTextJsonSerializer.cs
+++ b/src/JWT/Serializers/SystemTextJsonSerializer.cs
@@ -1,0 +1,60 @@
+﻿#if SYSTEMTEXTJSON
+namespace JWT.Serializers
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
+    using JWT;
+
+    public class SystemTextJsonSerializer : IJsonSerializer
+    {
+        public string Serialize(object obj)
+        {
+            // Implement using favorite JSON serializer
+            return JsonSerializer.Serialize(obj);
+        }
+
+        public T Deserialize<T>(string json)
+        {
+            // Implement using favorite JSON serializer
+            var data = JsonSerializer.Deserialize<T>(json);
+
+            if (data is Dictionary<string, object> odata)
+            {
+                var ndata = new Dictionary<string, object>();
+                foreach (var key in odata.Keys)
+                {
+                    var value = (JsonElement)odata[key];
+                    switch (value.ValueKind)
+                    {
+                        case JsonValueKind.Undefined:
+                            break;
+                        case JsonValueKind.Object:
+                            break;
+                        case JsonValueKind.Array:
+                            break;
+                        case JsonValueKind.String:
+                            ndata.Add(key, value.GetString());
+                            break;
+                        case JsonValueKind.Number:
+                            ndata.Add(key, value.GetInt64());
+                            break;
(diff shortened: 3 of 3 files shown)


src/JWT/JWT.csproj
```diff
@@ -36,8 +36,16 @@
...
+    <DefineConstants>SYSTEMTEXTJSON;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
+  </ItemGroup>
+
   <ItemGroup>
-   <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
```
abatishchev, line 48: Should we reference either one or another?

src/JWT/Builder/JwtHeader.cs
```diff
@@ -8,24 +11,45 @@ namespace JWT.Builder
     public class JwtHeader
     {
         [JsonProperty("typ")]
+#if SYSTEMTEXTJSON
```
abatishchev, line 14: Should it be one or another (conditionally)?

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            if (data is Dictionary<string, object> odata)
+            {
+                var ndata = new Dictionary<string, object>();
+                foreach (var key in odata.Keys)
+                {
+                    var value = (JsonElement)odata[key];
+                    switch (value.ValueKind)
+                    {
+                        case JsonValueKind.Undefined:
```
abatishchev, line 30: Please group all cases that do nothing into a single block

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+                            break;
+                        case JsonValueKind.Object:
+                            break;
+                        case JsonValueKind.Array:
+                            break;
+                        case JsonValueKind.String:
+                            ndata.Add(key, value.GetString());
```
abatishchev, line 37: And same here: multiple cases - one block of code

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+                            break;
+                        case JsonValueKind.False:
+                            ndata.Add(key, false);
+                            break;
+                        case JsonValueKind.Null:
+                            break;
+                    }
+                }
+
+                if (ndata is T obj) return obj;
```
abatishchev, line 53: Line break

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+            // Implement using favorite JSON serializer
+            return JsonSerializer.Serialize(obj);
+        }
+
+        public T Deserialize<T>(string json)
+        {
+            // Implement using favorite JSON serializer
+            var data = JsonSerializer.Deserialize<T>(json);
+
+            if (data is Dictionary<string, object> odata)
```
abatishchev, line 22: Invert the if?

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
+﻿#if SYSTEMTEXTJSON
+namespace JWT.Serializers
+{
+    using System;
```
abatishchev, line 4: usings outside the namespace, please

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+        public string Serialize(object obj)
+        {
+            // Implement using favorite JSON serializer
+            return JsonSerializer.Serialize(obj);
+        }
+
+        public T Deserialize<T>(string json)
+        {
+            // Implement using favorite JSON serializer
+            var data = JsonSerializer.Deserialize<T>(json);
```
abatishchev, line 20: Can you explain what this comment means?

src/JWT/JWT.csproj
```diff
@@ -36,8 +36,16 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
abatishchev, line 39: Please bump the version to 8.0.0-beta1

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
+﻿#if SYSTEMTEXTJSON
+namespace JWT.Serializers
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
+    using JWT;
+
+    public class SystemTextJsonSerializer : IJsonSerializer
```
abatishchev, line 9: Sealed

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public T Deserialize<T>(string json)
+        {
+            // Implement using favorite JSON serializer
+            var data = JsonSerializer.Deserialize<T>(json);
+
+            if (data is Dictionary<string, object> odata)
+            {
+                var ndata = new Dictionary<string, object>();
```
abatishchev, line 24: Why this code is needed at all?

src/JWT/Serializers/SystemTextJsonSerializer.cs
abatishchev: I think the another, already existing serializer must be included only if the target framework is lower than netstd2

src/JWT/JWT.csproj
abatishchev: And the another, already existing serializer must be included only if the target framework is lower than netstd2

src/JWT/Builder/JwtHeader.cs
luposky: I really don't know .. freedom of choiche is a value ... using netstandard20 I can choose beetween the two types of serializers (like in the asp.net core framework) so why forcing to use System.Text.Json if even microsoft don't force us to do so?

src/JWT/JWT.csproj
luposky: I don't see any problem in referencing Newtonsoft.Json in netstandard20 ... 
I simply like the option to use System.Text.Json ... 
maybe in the future with net50 things can be different and we can drop support for Newtonsoft.Json

src/JWT/Builder/JwtHeader.cs
abatishchev: Otherwise "replacing" won't be replacing but rather adding one more dependency, what users of Json.Net won't need hence would rightfully question.
Ideally either serializer would need to go to its own satellite package so users would be free to choose between one or another or maybe something third.
My point is that having both attributes and serializers basically means referencing them all and piling their respective attributes on each of this property.
I can easily imagine Unity folks or someone on .NET 3.5 who's unable to use either and who use DataContractJsonSerializer instead coming here and asking to add a third one.
So instead I would switch from a 3rd party to the 1st party serializer (and hope most of the users would be fine with that).

src/JWT/Serializers/SystemTextJsonSerializer.cs
luposky: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1200.md

src/JWT/Builder/JwtHeader.cs
luposky: yes, I understand ... I leave this choiche to you ... I push another commit with some improvment ...

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,59 @@
...
+                    case JsonValueKind.Undefined:
+                    case JsonValueKind.Object:
+                    case JsonValueKind.Array:
+                    case JsonValueKind.Null:
+                    default:
+                        throw new NotSupportedException(nameof(value.ValueKind));
+                }
+            }
+
+            return ndata is T obj ? obj : data;
```
abatishchev, line 54: Can you please explain this line?

src/JWT/Serializers/SystemTextJsonSerializer.cs
```diff
@@ -0,0 +1,58 @@
...
+            JsonSerializer.Serialize(obj);
+
+        public T Deserialize<T>(string json)
+        {
+            var data = JsonSerializer.Deserialize<T>(json);
+
+            // when deserializing a Dictionary<string, object>
+            // System.Text.Json create JsonElement objects for every value of the dictionary
+            // but application will expect to have native basic types (not a JsonElement class)
+            if (!(data is Dictionary<string, object> odata))
```
abatishchev, line 21: Do you know how it'll work if the type is IDictionary or IReadOnlyDictionary?

src/JWT/Serializers/SystemTextJsonSerializer.cs
luposky: You cannot "return ndata" because ndata is an object of type Dictionary<string, object> not of type T. This expression cast ndata to the type T if possible. If it is not possible, default to return the object previuosly deserialized as is. Compiler know nothing about the relationship between T and Dictionary<string, object>, so this trick (or similar) is required.

src/JWT/Serializers/SystemTextJsonSerializer.cs
luposky: I suppose it work out of the box with IDictionary (the concrete type returned by JsonSerializer.Deserialize<**IDictionary**<string, object>> is **Dictionary**<string, object>)
As I understand, the same happens with IReadOnlyDictionary, the concrete type deserialized is always a **Dictionary**<string, object>.

abatishchev: Great! Thanks for the clarification. I'll add tests once we decide how to handle this in general.

abatishchev: Hmm, I see. But direct cast to T won't work likely. Would `(T)(object)ndata` work here little better? Maybe the effect is same but more readable.

abatishchev:
What would you say if I propose to move the whole thing to a `Jwt.Serializers.SystemTextJson` package? Which would have its own version of the header model, decorated accordingly. You won't be able to use one of the DecodeHeader() methods but will be able the another.

Ideally I'd Invision to move all Json.Net-dependent classes to a separate package as well. This idea was outline long time ago in #

src/JWT/Serializers/SystemTextJsonSerializer.cs
luposky: Yes, it works, but the effect is not exactly the same ... if the double conversion don't work, null will be returned.

`return (T)(object)ndata ?? data` is the equivalent statement but I don't know if it is more readable ... I don't like double casting ... 

`if (ndata is T obj) return obj; else return data;` is another, more human readable, option ...

Issue #284, opened 2020-05-20
Custom 'typ' header


Altafard:
Will it be helpful if
```c#
header.Add("typ", "JWT");
```
will be replaced with
```c#
if (!header.ContainsKey("typ"))
    header.Add("typ", "JWT");
```

Are there any reasons not to do that?

abatishchev:
Do you mean in [this line](https://github.com/jwt-dotnet/jwt/blob/88cc93d81b9aa976e6cfcd6aec482039a06922e9/src/JWT/JwtEncoder.cs#L43)? I think it's a valid change. Please submit a PR! And bump the version slightly.

PR #285, opened 2020-05-21
Custom typ header

Altafard:
https://github.com/jwt-dotnet/jwt/issues/284

JwtEncoder checks if 'typ' header is presented in an extra headers before adding default value


abatishchev:
Cool! Almost done. Can you please add a simple test?

abatishchev:
Thanks much for the contribution!

abatishchev:
Published to NuGet as [7.2.1](https://www.nuget.org/packages/JWT/7.2.1) (might not be available right away, still indexing).

Issue #304, opened 2020-12-21
Add support for ECDSA signing algorithms


mandrachek:
java-jwt has out of the box support for RSA384, RSA512, and ECDSA256, ECDSA384, and ECDA512.

abatishchev:
Hi,
For RSAxyz please see my response at https://github.com/jwt-dotnet/jwt/issues/238, correct me if I'm wrong.

For ECDSAxyz some simple code would be needed, from what I can see at

abatishchev:
Sorry, pressed wrong button, on mobile they're so close to each other.

If you'd like to see this new functionality then please start a PR and I'll be more than happy to get it reviewed, polished if necessary, and released asap.

mandrachek:
Sorry, I'm not a .net developer - I'm a primarily a java dev, who happens to be working with some folks that will need to implement JWT as well, and was hoping I could refer them to this library. I found this to be a shortcoming when comparing the implementations.  Last C# code I wrote was about 10 years go. :/

#238 is actually incorrect. RSA256 uses a SHA-256 _hash_.  While it allows for any size RSA key, it doesn't allow for any size hashing algorithm (SHA-384, SHA-512).  It should be relatively straightforward to take the existing RS256 class, convert it into a an abstract base RSA class, and then RSA256, RSA384, and RSA512 concrete classes which just use a different HashAlgorithmName. Or something along those lines.

The EC based values are different in that they crank both the elliptic curve value, along with the hashing algorithm.

simon-pearson:
@abatishchev I'd like to take a look at this.

Would you object to me introducing a `RSASSAAlgorithm` abstract base class and inheriting from this in `RS256Algorithm`/`RS384Algorithm`/RS512Algorithm`?  Most of the code currently in `RS256Algorithm` would be moved into this class, and we'd add an abstract property for the hashing algorithm which each subclass could implement.

abatishchev:
@simon-pearson, sure. please! sounds like the way to go. i'd happy to participate, review the code, get it rolling.

simon-pearson:
@abatishchev unfortunately I think implementing the ECDSA algorithms will be much harder.  The `ECDsaCertificateExtensions.GetECDsaPrivateKey(X509Certificate2)` method used in https://stackoverflow.com/a/38135590/41956 will do the trick, but it was only added in .NET Framework 4.6.1/.NET Standard 1.4.

[...]

PR #309, opened 2021-02-06
Add support for ECDSA signing algorithms

simon-pearson:
An initial stab at #304.

A few points of discussion:
- The implementation of the `ECDSAAlgorithm` class relies heavily on the [`ECDsa.SignData`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.signdata?view=net-5.0), [`ECDsa.VerifyData`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.verifydata?view=net-5.0), [`ECDsaCertificateExtensions.GetECDsaPublicKey()`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsapublickey) and [`ECDsaCertificateExtensions.GetECDsaPrivateKey()`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsaprivatekey) methods, all of which were introduced in .NET Standard 1.4.  Our only target >= .NET Standard 1.4 is .NET Standard 2.0, so I've excluded this class from the build in all targets except .NET Standard 2.0.
- If the `ECDSAAlgorithmFactory.Create` method is called in any target other than .NET Standard 2.0 it throws `NotImplementedException`.
- Possibly we could re-work the preprocessor directives to include the `ECDSAAlgorithm`/`ES256Algorithm`/`ES384Algorithm`/`ES512Algorithm` classes in other targets but make them throw if you call any public methods on them.  Is there merit to this?

As always this is just a starter, feel free to chop + change as you see fit 👍


abatishchev (approved):

abatishchev:
@simon-pearson awesome!! i've bumped the version to `8.1.0-beta1` and ready to ship it.

abatishchev:
Just an idea for a separate PR, to improve tests: right now there is a test class for each algorithm and they all are identical in other than the algorithm type. We can leverage MSTest2's `[DataTestMethod]`:

```csharp
[DynamicData(nameof(GetData), DynamicDataSourceType.Method)]
[DataTestMethod]
public void Test(string str, Foo foo)
{
}

// this signature must stay
private static IEnumerable<object[]> GetData()
{
    // the number of objects must match the parameters
    yield return new object[]
    {
        "foo",
        new Foo()
    };
    yield return new object[]
    {
        "bar",
        new Bar()
    };
}
```
See if the same set of tests can accept different algorithm objects?

abatishchev:
Released as [8.1.0-beta1](https://www.nuget.org/packages/JWT/8.1.0-beta1). Let's promote it to a release in a week or two, if there will be no complains.

abatishchev:
Thanks you so much for both contributions!! You've made a huge difference in a very short period of time.

k15tfu:
> * The implementation of the `ECDSAAlgorithm` class relies heavily on the [`ECDsa.SignData`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.signdata?view=net-5.0), [`ECDsa.VerifyData`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.verifydata?view=net-5.0), [`ECDsaCertificateExtensions.GetECDsaPublicKey()`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsapublickey) and [`ECDsaCertificateExtensions.GetECDsaPrivateKey()`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsaprivatekey) methods, all of which were introduced in .NET Standard 1.4.  Our only target >= .NET Standard 1.4 is .NET Standard 2.0, so I've excluded this class from the build in all targets except .NET Standard 2.0.

@simon-pearson Hi! Just curious how did you check it? As far as I can see all of these four methods
* [`ECDsa.SignData(Byte[], HashAlgorithmName)`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.signdata?view=netframework-4.6.1#system-security-cryptography-ecdsa-signdata(system-byte()-system-security-cryptography-hashalgorithmname)) ,
* [`ECDsa.VerifyData(Byte[], Byte[], HashAlgorithmName)`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.verifydata?view=netframework-4.6.1#system-security-cryptography-ecdsa-verifydata(system-byte()-system-byte()-system-security-cryptography-hashalgorithmname)) ,
* [`ECDsaCertificateExtensions.GetECDsaPublicKey(X509Certificate2)`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsapublickey?view=netframework-4.6.1) ,
* and [`ECDsaCertificateExtensions.GetECDsaPrivateKey(X509Certificate2)`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.ecdsacertificateextensions.getecdsaprivatekey?view=netframework-4.6.1)


were introduced in .NET Framework 4.6.1 (see links), but not .NET Standard 1.4. Probably it worth considering upgrading the minimum target framework to net461 (or even newer) in the project, because otherwise these algorithms will not be available on the latest versions of the .NET Framework.

PR #315, opened 2021-02-24
Added support for .NET 5

### LINKED ISSUE #308: Build & test pipeline needs to support .NET 5
Right now it tests only 2.2 and 3.1.

abatishchev:
Resolves #308


abatishchev:
[Build ](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=664&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=006e6f7c-8314-5e02-a3df-483f3a1f4cfa&l=16)fails with error:

>##[error]C:\hostedtoolcache\windows\dotnet\sdk\5.0.103\Microsoft.Common.CurrentVersion.targets(1180,5): Error MSB3644: The reference assemblies for .NETFramework,Version=v3.5 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks

Whatever that means.

lechu445:
@abatishchev error during build
```
tests\JWT.Tests.Common\JwtValidatorTests.cs(10,27): Error CS0122: 'EncodingHelper' is inaccessible due to its protection level
```
suggests that internal classes are not visible in `JWT.Tests.Net50` project.

Adding line 
```c#
[assembly: InternalsVisibleTo("JWT.Tests.Net50, PublicKey=" + AssemblyInfo.PublicKey)]
```
into `jwt/src/JWT/Properties` should fix it.

abatishchev:
@lechu445 thank you!! completely forgot about this file, that it needs to list all test projects one by one.

abatishchev:
No, same outcome( As soon as I install .NET 5 on the build agent - it stops seeing .NET 3.5 targeting pack.

src/JWT/JWT.csproj
```diff
@@ -1,26 +1,28 @@
...
   <PropertyGroup>
     <Copyright>Public Domain</Copyright>
     <PackageDescription>Jwt.Net, a JWT (JSON Web Token) implementation for .NET</PackageDescription>
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>8.1.5</Version>
-    <FileVersion>8.0.0.0</FileVersion>
-    <AssemblyVersion>8.0.0.0</AssemblyVersion>
+    <Version>9.0.0</Version>
```
lechu445, line 23: The change does not affect public API and does not change behavior of the library. Change of version 8.1.5 to 9.0.0 suggests that this is a breaking change.  

I would suggest to not change version at all, as this change is only test project related. There is no change from library consumer point of view.

abatishchev: SemVer allows bumping the major version for major changes too, not necessarily only breaking changes in the public API. You might be right though, it can be the minor version just fine.

abatishchev: >I would suggest to not change version at all, as this change is only test project related. There is no change from library consumer point of view.

There is: it adds `net50` build target. So it's not just a test project that is changed.

lechu445:
@abatishchev this might be related to the issue https://github.com/dotnet/msbuild/issues/5985

abatishchev:
Published to nuget.org as [8.2.0](https://www.nuget.org/packages/JWT/8.2.0).

PR #316, opened 2021-02-28
Decoupled claim validation from JwtValidator

muhamad:
The `JwtValidator` has static validation for `exp` and `nbf` claims hard coded and in specific order which prevent the user from the ability to add a new behavior to the validation process.

Here I've created the type `JwtClaimValidation` which allow validating any JWT claim except the aud - for now - and moved the code for validating the `exp` and `nbf` to there, allowing either using these functions or supply others for custom scenarios.

Also modified the unit tests to support the new type.

The only breaking changes here - if acceptable - are the validation for `exp`  and `nbf` can be optional or mandatory for either of them.


abatishchev:
Hi @muhamad, thanks for your contribution! I need a little more time to understand and review the change.

Meanwhile I think what you'd like to achieve can be achieved without a change in the library by using the decorator pattern:

1. Implement the validator interface with your own class
2. Accept the validator interface into that class' ctor
3. Pass the default validator implementation as the latter to the ctor
4. Pass your validator implementation as the former to the builder

If you're using Microsoft DI container then I'd recommend to use [Scrutor](https://www.nuget.org/packages/Scrutor/), Simple Injector supports this pattern [out of the box.

abatishchev:
Regarding the change, I see the gated build has failed. Please make some changes:

1. I see [Directory.Build.props](https://github.com/jwt-dotnet/jwt/blob/main/Directory.Build.props#L4) already LangVersion=latest but still lambda discard doesn't work.
2. Bump the version in JWT.csproj

muhamad:
I've replaced the discard parameters in lambda expression to named ones, it seems the latest compiler version that is used in the pipeline don't support it.

Also updated `JwtClaimValidation` instance in the unit test to have input match the original behavior.

abatishchev:
Hey! Thanks for coming back to the PR and making the changes I asked for. I'm happy to stay engaged on this issue and get it through.

But first I'd like to better understand the rationale behind it. Why it's necessary to modify the built-in implementation of the interface rather than to implement it the way you'd like, and optionally - pick up the existing implementation: you can invoke it before _or_ after your custom business logic.

abatishchev:
Btw you can update the readme file directly in your PR. But I think the change should be made in a way to preserver the existing behavior with a possibility to change it, not the other way around.

muhamad:
> But first I'd like to better understand the rationale behind it. Why it's necessary to modify the built-in implementation of the interface rather than to implement it the way you'd like, and optionally - pick up the existing implementation: you can invoke it before or after your custom business logic.

The problem in current implementation is the signature verification implicitly validate claims which - IMHO - they should be separated because claims are not part of the signature but the payload.

The current implementation makes it difficult to achieve that and decorating current implementation also fails because the verification process is implicitly doing both. My implementation is the original with small modification - hence the PR 😃 - with slight difference.

Also in the AspNetCore Middleware we will need more options to validate some claims like `aud`, `iss` and so on, so I thought if the claim validation is separated from current implementation it will be good.

Another way of doing it without breaking the current implementation is to break current implementation into two classes:
  1. the `JwtValidatorBase` which have the basic pipeline for validation.
  2. the `JwtValidator` which have the original validation.

This way we can have multiple implementation which share the core logic and will allow extending the library without having to copy any code around.

> But I think the change should be made in a way to preserver the existing behavior with a possibility to change it, not the other way around.
Agreed.

I think we should preserve the current behavior as it is but also make it more flexible to extend instead of reimplementing it from scratch to add small features.

I know my implementation may not suits all, So if you consider the above approach I will be happy to implement it, or any other approach we will agree upon.

muhamad:
a better implementation in #320

PR #320, opened 2021-03-01
Implemented another approach to claim validation

muhamad:
As per #316 , I've re-implemented the claim validation taking another approach which don't require any client code change and it only update exception messages in unit tests - 3 tests only.

In this approach I've simplified the design greatly to make it more friendly. the only change is in the interface `IJwtValidator` which now have a new property `ClaimValidation` to allow updating its members from outside the validator, e.g. from validation handler.


abatishchev:
Hi @muhamad, I'm sorry? The email notification about your PR felt through the cracks, I missed it.

In general, I'd be against the changes in the interface. Especial of you want to change its implementation.

Can you please explain how exactly you want to change the current flow? Do you want to substitute it, or run additional code before or after the default?

If the latter then the decorator pattern would work here the best, I can help with some samples.

If the former then the default validator can optionally accept an interface do claims validation. And by default would supply the default implementation to provide the existing, default behavior.

PR #325, opened 2021-04-07
Memory optimization in JwtEncoder.Encode method

lechu445:
It reduces a few unneeded allocations.

### DIFF
--- a/src/JWT/JwtEncoder.cs
+++ b/src/JWT/JwtEncoder.cs
@@ -37,29 +37,26 @@ namespace JWT
             if (!_algorithm.IsAsymmetric() && key is null)
                 throw new ArgumentNullException(nameof(key));
 
-            var segments = new List<string>(3);
-
             var header = extraHeaders is null ?
-                new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase) :
+                new Dictionary<string, object>(2, StringComparer.OrdinalIgnoreCase) :
                 new Dictionary<string, object>(extraHeaders, StringComparer.OrdinalIgnoreCase);
             
-            if (!header.ContainsKey("typ"))
-                header.Add("typ", "JWT");
+            header.TryAdd("typ", "JWT");
             header.Add("alg", _algorithm.Name);
 
             var headerBytes = GetBytes(_jsonSerializer.Serialize(header));
             var payloadBytes = GetBytes(_jsonSerializer.Serialize(payload));
 
-            segments.Add(_urlEncoder.Encode(headerBytes));
-            segments.Add(_urlEncoder.Encode(payloadBytes));
+            var headerSegment = _urlEncoder.Encode(headerBytes);
+            var payloadSegment = _urlEncoder.Encode(payloadBytes);
 
-            var stringToSign = String.Join(".", segments.ToArray());
+            var stringToSign = headerSegment + "." + payloadSegment;
             var bytesToSign = GetBytes(stringToSign);
 
             var signature = _algorithm.Sign(key, bytesToSign);
-            segments.Add(_urlEncoder.Encode(signature));
+            var signatureSegment = _urlEncoder.Encode(signature);
 
-            return String.Join(".", segments.ToArray());
+            return stringToSign + "." + signatureSegment;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/JWT/JwtEncoder.cs
```diff
@@ -50,16 +48,16 @@ public string Encode(IDictionary<string, object> extraHeaders, object payload, b
             var headerBytes = GetBytes(_jsonSerializer.Serialize(header));
             var payloadBytes = GetBytes(_jsonSerializer.Serialize(payload));
 
-            segments.Add(_urlEncoder.Encode(headerBytes));
-            segments.Add(_urlEncoder.Encode(payloadBytes));
+            var headerSegment = _urlEncoder.Encode(headerBytes);
+            var payloadSegment = _urlEncoder.Encode(payloadBytes);
 
-            var stringToSign = String.Join(".", segments.ToArray());
+            var stringToSign = headerSegment + "." + payloadSegment;
```
abatishchev, line 54: Looking at this article https://dotnetcoretutorials.com/2020/02/06/performance-of-string-concatenation-in-c/ I see the author claims that Join and StringBuilder are the fastest. So would this work the best:

```c#
var stringToSign = String.Join('.', headerSegment, payloadSegment);
```
?

src/JWT/JwtEncoder.cs
```diff
@@ -50,16 +48,16 @@ public string Encode(IDictionary<string, object> extraHeaders, object payload, b
...
-            var stringToSign = String.Join(".", segments.ToArray());
+            var stringToSign = headerSegment + "." + payloadSegment;
             var bytesToSign = GetBytes(stringToSign);
 
             var signature = _algorithm.Sign(key, bytesToSign);
-            segments.Add(_urlEncoder.Encode(signature));
+            var signatureSegment = _urlEncoder.Encode(signature);
 
-            return String.Join(".", segments.ToArray());
+            return stringToSign + "." + signatureSegment;
```
abatishchev, line 60: And same here?

abatishchev:
Please bump the version in `JWT.csproj`.

lechu445:
changed version of JWT.csproj from 8.1.4 to 8.1.5

src/JWT/JwtEncoder.cs
lechu445: The author of the arcticle did not consider our use-case where we concatenate two strings with separator. Also he did not measure memory allocation.

I did better suited benchmark test fot this use-case
```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.201
[Host]     : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT
DefaultJob : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT

|            Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------ |---------:|---------:|---------:|-------:|------:|------:|----------:|
| StringConcatenate | 18.65 ns | 0.068 ns | 0.056 ns | 0.0089 |     - |     - |      56 B |
|        StringJoin | 29.98 ns | 0.578 ns | 0.541 ns | 0.0153 |     - |     - |      96 B |
```

benchmark test code
```c#
using BenchmarkDotNet.Attributes;

namespace Benchmarks
{
  [MemoryDiagnoser]
  public class StringJoinBenchmark
  {
    public string string1 = "testing1";
    public string separator = ".";
    public string string2 = "testing2";

    [Benchmark]
    public string StringConcatenate() => string1 + " " + string2;

    [Benchmark]
    public string StringJoin() => string.Join(" ", string1, string2);
  }
}
```

src/JWT/JwtEncoder.cs
lechu445: same answer as above

src/JWT/JwtEncoder.cs
lechu445: the reason why it is faster now is that compiler compiles `string1 + " " + string2` as `string.Concat(string1, " ", string2)`. And string.Concat has special case overload that takes three strings as parameters.  So, it is optimized for this use-case. https://source.dot.net/#System.Private.CoreLib/String.Manipulation.cs,17e1cc2f91f20974

abatishchev (approved):

abatishchev:
Perfect. Thanks a lot for the contribution! I'll publish to nuget shortly.

PR #326, opened 2021-04-09
Various performance and memory allocation optimizations

lechu445:
- changed call ToString() on enums, and used nameof as this produce const string at compilation time, so const can be processed by JIT in a more efficient way
- reused result of `value.ToString()` in EnumExtensions.GetDescription method
- reduced allocation of unnecessary string.Split result array in JwtData and JwtBase64UrlEncoder
- reused UTF8Encoding instance in EncodingHelper, as I believe it won't change after first usage
- introduced a new method GetBytes in EncodingHelper that allows join two strings directly into bytes array
- reduced Linq usage where it could be replaced with type-specific methods with equivalent behaviour
- reduced yields by joining two Linq Select calls in JwtDecoder.Validate method
- minor comments fixes
- changed version of JWT project to 8.1.6

### DIFF
--- a/src/JWT/Builder/EnumExtensions.cs
+++ b/src/JWT/Builder/EnumExtensions.cs
@@ -24,9 +24,9 @@ namespace JWT.Builder
         {
             var valueString = value.ToString();
             return value.GetType()
-                .GetField(valueString)
-                .GetCustomAttribute<DescriptionAttribute>()
-                ?.Description ?? valueString;
+                        .GetField(valueString)
+                        .GetCustomAttribute<DescriptionAttribute>()
+                       ?.Description ?? valueString;
         }
     }
 }
--- a/src/JWT/Builder/EnumExtensions.cs
+++ b/src/JWT/Builder/EnumExtensions.cs
@@ -20,10 +20,13 @@ namespace JWT.Builder
         /// <summary>
         /// Gets the value of the <see cref="DescriptionAttribute" /> from the object.
         /// </summary>
-        private static string GetDescription<T>(T value) =>
-            value.GetType()
-                 .GetField(value.ToString())
-                 .GetCustomAttribute<DescriptionAttribute>()
-                ?.Description ?? value.ToString();
+        private static string GetDescription<T>(T value)
+        {
+            var valueString = value.ToString();
+            return value.GetType()
+                .GetField(valueString)
+                .GetCustomAttribute<DescriptionAttribute>()
+                ?.Description ?? valueString;
+        }
     }
 }
--- a/src/JWT/Compatibility/String.cs
+++ b/src/JWT/Compatibility/String.cs
@@ -1,16 +1,21 @@
-#if NET35
 using System;
 
 namespace JWT.Compatibility
 {
     internal static class String
     {
+#if !NETSTANDARD
+        /// <remarks>
+        /// The overload accpting char is missing outside .NET Standard.
+        /// See https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=netstandard-2.0#System_String_IndexOf_System_String_System_StringComparison_
+        /// </remarks>
         public static int IndexOf(this string value, char ch, StringComparison comparisonType)
         {
-            // ignores comparisonType since the overload is missing in .NET 3.5
-            return value.IndexOf(ch);
+            return value.IndexOf(ch.ToString(), comparisonType);
         }
-    
+#end
+  
+#if NET35
         public static bool IsNullOrWhiteSpace(string value)
         {
             if (!string.IsNullOrEmpty(value))
@@ -23,5 +28,5 @@ namespace JWT.Compatibility
             return true;
         }
     }
-}
 #endif
+}
--- a/src/JWT/Compatibility/String.cs
+++ b/src/JWT/Compatibility/String.cs
@@ -5,7 +5,7 @@ namespace JWT.Compatibility
 {
     internal static class String
     {
-        public int IndexOf(string value, char ch, StringComparison comparisonType)
+        public int IndexOf(this string value, char ch, StringComparison comparisonType)
         {
             // ignores comparisonType since the overload is missing in .NET 3.5
             return value.IndexOf(ch);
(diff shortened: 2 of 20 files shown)


abatishchev (approved):
Took the liberty to make few additional changes. Can you please take a look to confirm they're ok?

src/JWT/Builder/EnumExtensions.cs
```diff
@@ -20,10 +20,13 @@ internal static class EnumExtensions
         /// <summary>
         /// Gets the value of the <see cref="DescriptionAttribute" /> from the object.
         /// </summary>
-        private static string GetDescription<T>(T value) =>
-            value.GetType()
-                 .GetField(value.ToString())
-                 .GetCustomAttribute<DescriptionAttribute>()
-                ?.Description ?? value.ToString();
+        private static string GetDescription<T>(T value)
+        {
```
abatishchev, line 24: Can we take the advantage of relatively-new constraint `where T : enum` here and do something more optimal?

abatishchev (approved):

src/JWT/Compatibility/String.cs
```diff
@@ -1,8 +1,16 @@
...
+using System;
+
 namespace JWT.Compatibility
 {
     internal static class String
     {
+        public int IndexOf(string value, char ch, StringComparison comparisonType)
+        {
+            // ignores comparisonType since the overload is missing in .NET 3.5
+            return value.IndexOf(ch);
```
abatishchev, line 11: Would just ignoring `comparisonType` be ok? Or it's better to use [the another overload](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=netframework-3.5#System_String_IndexOf_System_String_System_StringComparison_):

```c#
public int IndexOf (string value, StringComparison comparisonType);
```
i.e.:

```c#
return value.IndexOf(ch.ToString(), comparisonType);
```

abatishchev (approved):

src/JWT/Compatibility/String.cs
abatishchev: I give up trying to make this working (

src/JWT/JWT.csproj
```diff
@@ -18,7 +18,7 @@
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>8.1.5</Version>
+    <Version>8.2.0</Version>
```
lechu445, line 21: This PR does not add a new feature. I would suggest to change only PATCH part of the version because it does not change the public API.

I think it is fine if user defines following in his csproj:
```xml
<PackageReference Include="JWT" Version="8.1.*" />
```
nuget resolves the optimized version.

```suggestion
    <Version>8.1.6</Version>
```

src/JWT/Builder/EnumExtensions.cs
lechu445: Short answer: there is no benefit of the proposed change

Long answer:
`where T : enum` does not compile. I checked two options:
- `where T : Enum`
- `where T : struct`

results of benchmarks:
```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.201
[Host]     : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT
DefaultJob : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT

|             Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|            Current | 984.4 ns | 16.97 ns | 15.88 ns | 0.0381 |     - |     - |     248 B |
|   Enum_Constrained | 990.9 ns | 18.60 ns | 16.49 ns | 0.0381 |     - |     - |     248 B |
| Struct_Constrained | 983.2 ns | 16.67 ns | 15.59 ns | 0.0381 |     - |     - |     248 B |
```
I used the newest version of .NET Core to test because it has the most optimisation hacks in Roslyn/JIT.

As you can see, the results are pretty the same.  
Also IL code built in Release mode does not have important differences.

Here is benchmark tests that I used to compare the solutions:
```c#
  [MemoryDiagnoser]
  public class EnumDescriptionBenchmark
  {
    [Benchmark]
    public string Current() => GetDescription(HeaderName.X5c);

    [Benchmark]
    public string Enum_Constrained() => GetDescription_Enum_Constrained(HeaderName.X5c);

    [Benchmark]
    public string Struct_Constrained() => GetDescription_Struct_Constrained(HeaderName.X5c);

    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private static string GetDescription<T>(T value) =>
      value.GetType()
           .GetField(value.ToString())
           .GetCustomAttribute<DescriptionAttribute>()
          ?.Description ?? value.ToString();

    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private static string GetDescription_Enum_Constrained<T>(T value) where T : Enum =>
      value.GetType()
           .GetField(value.ToString())
           .GetCustomAttribute<DescriptionAttribute>()
          ?.Description ?? value.ToString();

    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private static string GetDescription_Struct_Constrained<T>(T value) where T : struct =>
      value.GetType()
           .GetField(value.ToString())
           .GetCustomAttribute<DescriptionAttribute>()
          ?.Description ?? value.ToString();

    private enum HeaderName
    {
      [Description("typ")]
      Type,

      [Description("cty")]
      ContentType,

      [Description("alg")]
      Algorithm,

      [Description("kid")]
      KeyId,

      [Description("x5u")]
      X5u,

      [Description("x5c")]
      X5c,

      [Description("x5t")]
      X5t
    }
  }
```

src/JWT/JWT.csproj
abatishchev: I just wanted to encourage users to pick up the updated version faster. Ok, we can downgrade it back.

abatishchev:
Published to nuget.org [8.1.6](https://www.nuget.org/packages/JWT/8.1.6).

@lechu445 thank you so much for your contribution!

Issue #328, opened 2021-05-08
`RSAlgorithm.GetPrivateKey` and `RSAlgorithm.GetPublicKey` should use `X509Certificate2.GetRSA...Key` on .NET Framework 4.6 and higher


safesparrow:
The problem I was facing was an "Invalid algorithm exception" when signing the token, as described here:
https://stackoverflow.com/questions/7433074/why-am-i-getting-invalid-algorithm-specified-exception

This StackOverflow comment showed me the solution:
https://stackoverflow.com/questions/7433074/why-am-i-getting-invalid-algorithm-specified-exception#comment111629660_7433074

The problem is that these two methods in `RSAlgorithm` use compile-time conditions to decide how to extract public & private key from a certificate:
```
        private static RSA GetPrivateKey(X509Certificate2 cert)
        {
            if (cert is null)
                throw new ArgumentNullException(nameof(cert));

#if NETSTANDARD1_3
            return cert.GetRSAPrivateKey();
#else
            return (RSA)cert.PrivateKey;
#endif
        }

        private static RSA GetPublicKey(X509Certificate2 cert)
        {
            if (cert is null)
                throw new ArgumentNullException(nameof(cert));

#if NETSTANDARD1_3
            return cert.GetRSAPublicKey();
#else
            return (RSA)cert.PublicKey.Key;
#endif
        }
```
I'm using .NET Framework 4.6 and although the two `GetRSA...Key()` methods are available in .NET Framework 4.6 and higher:
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.rsacertificateextensions.getrsapublickey?view=net-5.0
they were not being picked up by the above code.

The workaround for me was to use the constructor that accepts public & private key.
Although it's a very quick workaround, it's not at all obvious how to find it.

Would it be possible to update the branching logic to support .NET Framework?
Based on https://docs.microsoft.com/en-us/dotnet/standard/frameworks , the NETSTANDARD1_3 is only set in NETSTANDARD assemblies, and not .NET Core or .NET Framework ones, so I imagine .NET Core would require handling as well.
I think that each target framework sets only a single compile-time variable, so the code would have to check for all future NETSTANDARD, NET & NETCORE versions as well.

abatishchev:
Hi @safesparrow, thanks for reporting this issue!

May I ask to submit a PR? I'll get it reviewed and will publish an updated version to NuGet shortly thereafter.

abatishchev:
[...]

PR #332, opened 2021-05-15
Use `X509Certificate2.GetRSAPublicKey` in all supported TargetFrameworks

safesparrow:
This addresses https://github.com/jwt-dotnet/jwt/issues/328 .

The existing preprocessor directive only uses `X509Certificate2.GetRSA...Key()` if targetting NETSTANDARD1_3.

Based on https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.rsacertificateextensions.getrsapublickey?view=net-5.0 , these two methods are supported everywhere except in NETFramework < 4.6

This PR changes the preprocessor branching to list all target frameworks that support `X509Certificate2.GetRSA...Key()`:
* Any netstandard, netcore or net versions
* netframework >= 4.6, that is: 46, 461, 462, 47, 471, 472, 48

### DIFF
--- a/src/JWT/Algorithms/RSAlgorithm.cs
+++ b/src/JWT/Algorithms/RSAlgorithm.cs
@@ -86,7 +86,7 @@ namespace JWT.Algorithms
             if (cert is null)
                 throw new ArgumentNullException(nameof(cert));
 
-#if NETSTANDARD1_3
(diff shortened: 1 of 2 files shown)


abatishchev (approved):

abatishchev:
@safesparrow awesome, thank you! hope this will work now. please try it out using [8.2.2](https://www.nuget.org/packages/JWT/8.2.2).

PR #333, opened 2021-05-19
Added token validation parameters

### LINKED ISSUE #331: Option to turn off or ignore certain token parameters during token validation
I wish to parse and validate some JWTs that have expired (to generate new tokens with a refresh token). However, when parsing a token using the following code:

```csharp
string json = JwtBuilder.Create()
    .WithSecret(secret)
    .MustVerifySignature()
    .Decode<string>(token);
```

an exception is thrown if the token has expired.

To workaround the issue, I have to change the code to this:

```csharp
IJsonSerializer serializer = new JsonNetSerializer();
UtcDateTimeProvider provider = new UtcDateTimeProvider();
IJwtValidator validator = new JwtValidator(serializer, provider);
IBase64UrlEncoder urlEncoder = new JwtBase64UrlEncoder();
IJwtAlgorithm algorithm = new HMACSHA256Algorithm();
JwtDecoder decoder = new JwtDecoder(serializer, validator, urlEncoder, algorithm);

JwtParts jwtParts = new JwtParts(token);

try
{
    decoder.Validate(jwtParts, secret);
}
catch (TokenExpiredException)
{
    // do nothing
}

string json = decoder.Decode(token);
```

which is quite uglier than the original fluent style code. The other option was to create a new validation class that inherits from `IJwtValidator` and remove the expiration validation, but this would create a lot of duplicate code.

Ideally, we would be able to pass options to the method `.MustVerifySignature()`, something like the following:

```csharp
string json = JwtBuilder.Create()
    .WithSecret(secret)
    .MustVerifySignature(options =>
        options.TokenValidationParameters = new TokenValidationParameters
        {
            ValidateIssuerSigningKey = true,
            ValidateIssuer = false,
            ValidateAudience = false,
            ValidateLifetime = true,
            ClockSkew = TimeSpan.Zero
        }
    )
    .Decode<string>(token);
```

This would allow us to control the execution of certain parts of the validation class to suit the individual needs of the token validation.

Let me know what you think, and I'd be happy to create a PR to address this issue.

nathanpovo:
Adds the option to pass validation parameters to the constructor of `JwtValidator` to turn off certain parts of the token validation.

The validation parameters can also be passed to the `WithVerifySignature` method to be used with the fluent API.

Fixes #331


src/JWT/JwtValidator.cs
```diff
@@ -28,28 +28,21 @@ public sealed class JwtValidator : IJwtValidator
         private readonly IJsonSerializer _jsonSerializer;
         private readonly IDateTimeProvider _dateTimeProvider;
         private readonly int _timeMargin;
+        private readonly ValidationParameters _validationParameters;
```
abatishchev, line 31: Maybe call it `valParams` or `validationParams`, just to shorten a little.

src/JWT/JwtValidator.cs
```diff
@@ -28,28 +28,21 @@ public sealed class JwtValidator : IJwtValidator
...
         private readonly IDateTimeProvider _dateTimeProvider;
         private readonly int _timeMargin;
+        private readonly ValidationParameters _validationParameters;
 
         /// <summary>
         /// Creates an instance of <see cref="JwtValidator" />
         /// </summary>
         /// <param name="jsonSerializer">The Json Serializer</param>
         /// <param name="dateTimeProvider">The DateTime Provider</param>
-        public JwtValidator(IJsonSerializer jsonSerializer, IDateTimeProvider dateTimeProvider)
```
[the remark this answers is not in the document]
nathanpovo: I mostly did that to not have to introduce two overloads to the constructor.

From [this](https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/) link it seems that even overloads might cause the library to be binary incompatible.

If you'd prefer that overloads were used, just let me know and I'll add them in.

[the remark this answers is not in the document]
abatishchev: Yeah, binary backward compatibility is a complex problem, and I'm not too familiar with it. But in general I'm rather old-school and feel that overloads are more readable than optional parameters. Let's have an overload here, if you'd be ok with that. Thanks!

abatishchev (approved):

abatishchev:
Thanks for adding tests! Everyone like to have them and nobody likes to write them :) In long-run that's the single most beneficial change among any. Please add as much tests as possible, see if you can add any additional tests.

abatishchev:
@nathanpovo sorry for coming late! I've made few changes? Can you please take a look?

Also I'm thinking `TimeMargin` must be a property on `ValidationParameters`. Because it's a validation parameter too.

Since this is your PR, I don't want to make too many changes myself. If you agree, can you please do that? The actor accepting `timeMargin` may just update to the value. Or we can remove it altogether, effectively making a breaking change and bump version to `9.0.0`.

src/JWT/ValidationParameters.cs
```diff
@@ -1,38 +1,48 @@
...
         /// Gets or sets a boolean to control if the lifetime will be validated during token validation.
         /// </summary>
-        public bool ValidateLifetime { get; set; } = true;
+        public bool ValidateLifetime { get; set; }
 
         /// <summary>
         /// Gets or sets a boolean to control if the issued time will be validated during token validation.
         /// </summary>
-        public bool ValidateIssuedTime { get; set; } = true;
+        public bool ValidateIssuedTime { get; set; }
```
nathanpovo: Could these default values be added back? Creating a new `ValidationParameters` object with a single property value set to `false` (such as in the tests) without these default values would require a consumer to either define the value for each property:
```csharp
var valParams = new ValidationParameters
{
    ValidateIssuerSigningKey = true,
    ValidateLifetime = true,
    ValidateIssuedTime = false
};
```
or use the default property `ValidationParameters.Default` and then manually set the value of the property you want to change:
```csharp
var valParams = ValidationParameters.Default;
valParams.ValidateIssuedTime = false;
```
which seems a bit clunky.

src/JWT/JwtValidator.cs
```diff
@@ -71,10 +71,10 @@ public JwtValidator(IJsonSerializer jsonSerializer, IDateTimeProvider dateTimePr
         /// <param name="validationParameters">Validation parameters that are passed on to <see cref="JwtValidator"/></param>
         public JwtValidator(IJsonSerializer jsonSerializer, IDateTimeProvider dateTimeProvider, int timeMargin, ValidationParameters valParams)
         {
-            _jsonSerializer = jsonSerializer;
-            _dateTimeProvider = dateTimeProvider;
+            _jsonSerializer = jsonSerializer ?? throw new ArgumentNullException(nameof(jsonSerializer));
+            _dateTimeProvider = dateTimeProvider ?? throw new ArgumentNullException(nameof(dateTimeProvider));
```
nathanpovo, line 75: The tests in `JwtValidatorTests` are failing due to this change; they are expecting an exception of type `ArgumentException` not `ArgumentNullException`.

I'm not sure if this is enough to be considered a breaking change (in which case the library version would have to be bumped to `9.0.0`).

nathanpovo:
@abatishchev, the library is yours; you know what best to change in it 😄

About `TimeMargin` being a validation parameter, I think it makes sense to include it in `ValidationParameters`. And removing the `TimeMargin` would make the constructor less confusing; all the validation parameters would be in the same place.

If you want to go ahead with the breaking change of throwing `ArgumentNullException` instead of `ArgumentException` then I think the we can remove the `JwtValidator` constructors with the `TimeMargin`. If not, the constructors could be marked as obsolete for now and then removed at a later date.

abatishchev (approved):

src/JWT/JwtValidator.cs
abatishchev: Oh, wow. Can you please update the tests, if not already?

And yes, I think it's fine to bump the version to 9.0.0, did that.

src/JWT/ValidationParameters.cs
abatishchev: Do we want to introduce Fluent API for ValidationParameters to? It can be simply a set of extensions methods:
```c#
public ValidationParameters NotValidateIssuedTime(this ValidationParameters  valParams) =>
    valParams.ValidateIssuedTime(false);

public ValidationParameters ValidateIssuedTime(this ValidationParameters  valParams, bool value)
{
valParams.ValidateIssuedTime = value;
return valParams;
}
```

src/JWT/JwtValidator.cs
nathanpovo: Tests have already been updated; they are all passing.

src/JWT/ValidationParameters.cs
nathanpovo: I think that adding a fluent api for the `ValidationParameters` is not needed since it's just a class with a bunch of properties; using the fluent api would end up needing more code and would increase maintenance if any additional parameters were added.

However, I do think it should be clear what the default property values are.

Do you think the default value of the boolean properties should be `false` rather than `true`?

abatishchev: Hey @binki, I'm sorry once again! I'm getting overloaded at work, coming back to your PR as soon as get a spare moment.

I was thinking more about the default values and setting them all to `true` is unusual and counterintuitive. Should we have flip all the properties so they all would have `false` by default so just `new` would be identical to `Default`? Same story for `margin` which is `0` by default already.

Also maybe we should add the ctor accepting `margin` back? And make it pass the value to the ctor of parameters

binki: @abatishchev I’m sure you meant to mention @nathanpovo (though GitHub thinks you shouldn’t mention him because he is the one who reported this issue). However, I still have opinions. Having initialization on properties is not strange for classes. Changing the parameter names to allow default values to make sense (e.g., having `DoNotValidateIssuerSigningKey` implicitly initialized to `false` instead of `ValidateIssuerSigningKey` initialized to `true`) is only needed for structs.

I think it would be most bad to make it easy to accidentally turn validation off entirely. So I’d prefer the properties to be initialized to what are considered good defaults.

If you want people to use `Default` instead, then you should make the constructor `private` so that no one can accidentally instantiate the class directly. This way, they would be forced to choose `None` or `Default` instead of being able to accidentally use `None` by calling the constructor.

binki: > manually set the value of the property you want to change:
> 
> ```cs
> var valParams = ValidationParameters.Default;
> valParams.ValidateIssuedTime = false;
> ```
> 
> which seems a bit clunky.

If you find this clunky, you can always define an extension method like the following if you ever want to make sure you have fully initialized an object before making it available to the current scope:

```
public static class ObjectExtensions {
  public static T With<T>(
    this T o,
    Action<T> action) {
    action(o);
    return o;
  }
}
```

nathanpovo: @abatishchev, no worries! The feature isn't urgently needed, I'm just adding it to the library to help out and to make it a bit easier to use.

@binki, thanks for the valuable input. I agree that it's important to not make it easy to turn off the validation by accident; that's why I had set the property values to `true` so that the validation would get done by default. Setting the constructor to `private` seems like the easiest thing to do; that way we could leave the defaults of the properties to whatever the default of the type is and also prevent a user from creating validation parameters that will not validate. We could then use an extension method, as suggested by @binki, to easily change the values from the default.

binki: Thanks, I am glad to see discussion is happening!

I think I must clarify/condition my suggestion to use an extension method just to make sure we are understanding each other ;-). I do not think it would be appropriate to add the exact extension method (`With<T>`) I suggested to this library. That is more of a personal coding style/choice. I define that in my own projects because sometimes I want to be able to stuff related things into an expression. For example, I might use it to initialize [`XmlWriterSettings`](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlwritersettings?view=net-5.0). It belongs in a utility class or package for use by those who subscribe to this coding style.

What would be more appropriate would probably be a “fluent” API as discussed earlier. This could be as simple as calling a callback (like with my `With<T>`, but just targeted to the `ValidationParameters` type instead of generic).

abatishchev: @binki sorry for involving you into the discussion outside your will :) and thanks for valuable input!
Totally agree that the most important aspect should be preventing accidentally turning off the validation, that would be disastrous.

abatishchev: @nathanpovo generic `With<T>` is too generic in my opinion. it's better to add few `ValidationParameters`-specific methods. This way they'll be scoped, discoverable, and readable. what you think?
Sorry for going back and forth on this!

nathanpovo: Yeah, a generic method wouldn't make sense in this case; I added an extension method just to modify `ValidationParameters` objects.

src/JWT/ValidationParameters.cs
```diff
@@ -67,4 +69,13 @@ private ValidationParameters()
             TimeMargin = 0
         };
     }
+
+    public static class ValidationParametersExtensions
+    {
+        public static ValidationParameters With(this ValidationParameters @this, Action<ValidationParameters> action)
```
binki, line 75: I think you need to use a different name than `With<T>()`. Something like `SetValidationParameters` might work. I wonder if @abatishchev wants to more fully embrace the Fluent API style, adding a method per setting such as:

* `EnableValidateIssuerSigningKey`/`DisableValidateIssuerSigningKey`
* `EnableValidateLifetime`/`DisableValidateLifetime`
* etc.…

Could you specify a bit more clearly/fully what is acceptable, @abatishchev ? Thanks!

nathanpovo: I used `With` to make it synonymous with the `with` statements for `records` in `C# 9.0` (which, sadly, cannot be used here). Maybe it could be changed to something like `WithParameters`, but the `With` method seems clear enough

```csharp
var validationParameters = ValidationParameters.Default
    .With(parameters =>
    {
        parameters.ValidateLifetime = false;
        parameters.TimeMargin = 1;
    });
```

With regards to fully embracing the fluent api style, I think it would add additional maintanence without a lot of gain since the class is just a parameter class; any fluent methods would not have enough to do, they would just toggle the values of boolean parameters.

abatishchev:
@nathanpovo my sincere apologizes that this PR took forever me to come back (high demanding job and burn out among other things).
In order to more easily work on the changes and get them shipped, I'm closing this PR off your branch and opening #376 off mine.

abatishchev:
Release to nuget.org as [JWT 9.0.0-beta2](https://www.nuget.org/packages/JWT/9.0.0-beta2)

PR #347, opened 2021-09-12
Added more overloads for AddJwt<TFactory>()

abatishchev:


abatishchev:
@frankely pushed [version 7.3.1](https://www.nuget.org/packages/JWT.Extensions.AspNetCore/7.3.1) which contains more overloads, please try that one.

PR #351, opened 2021-09-23
Allowed keys to be null for asymmetric algorithms

### LINKED ISSUE #350: What is the relationship between IAlgorithmFactory and JwtAuthenticationOptions.Keys?
The RS256Algorithm requires a key in the ctor, yet I allow validation against multiple public keys. I handle it via something like this:

```
class RSAFactory : IAlgorithmFactory
{
…
    public IJwtAlgorithm Create(JwtDecoderContext context)
    {
    // get cert via KeyId from a kvp of identifier : publickey in settings
    var cert = GetCertFromConfig(context.Header.KeyId);
    // Also throw and log if not found

    Return new RS2560Algorithm(cert);
    }
}
```

In `Startup.cs`:

```
  services.AddAuthentication(x =>
    {
        x.DefaultAuthenticateScheme = JwtAuthenticationDefaults.AuthenticationScheme;
        x.DefaultChallengeScheme = JwtAuthenticationDefaults.AuthenticationScheme;
    }).AddJwt<RSAFactory>(options =>
    {                
         options.VerifySignature = true;
         //  keys can't be nothing but doesn’t have to be the actual keys
         options.Keys = new string[] { " "};
    }
```

Since the algorithm requires the public key, it has to be built via the factory. The options.Keys property is ignored. Is this the intended usage? The code works but I wanted to make sure that I’m not missing something.

The drawback is that I have to throw inside the factory if the KeyId doesn't match the config.

abatishchev:
Resolves #350


abatishchev:
@wtjones can you please try out the drop generated by this PR's build and let me know? if it helps - I'll push to NuGet immediately after.

wtjones:
I tried 8.4.1, but even when not setting keys or setting an empty array I get an error. I think this [method](https://github.com/jwt-dotnet/jwt/blob/5f11052347f89dbd390e59401af9cf8a8dfec44b/src/JWT/JwtDecoder.cs#L143) needs to be adjusted as well

` options.Keys = new string[] {};`
```
info: JWT.JwtAuthenticationHandler[2]
      Error decoding JWT: Specified argument was out of the range of valid values. (Parameter 'keys'), returning failure
      System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'keys')
         at JWT.JwtDecoder.Decode(JwtParts jwt, Byte[][] keys, Boolean verify)
         at JWT.JwtDecoder.DecodeToObject[T](JwtParts jwt, Byte[][] keys, Boolean verify)
         at JWT.JwtDecoderExtensions.DecodeToObject[T](IJwtDecoder decoder, String token, Byte[][] keys, Boolean verify)
         at JWT.JwtDecoderExtensions.DecodeToObject[T](IJwtDecoder decoder, String token, String[] keys, Boolean verify)
         at JWT.JwtAuthenticationHandler.GetAuthenticationResult(String header)
```

abatishchev:
@wtjones I see, means there are more places where the validation needs to be relaxed. Made the changes, please try [this build](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=849&view=artifacts&pathAsName=false&type=publishedArtifacts) once it's done in few minutes.

wtjones:
I'm having some trouble ensuring that I have the correct package, but now I am getting this if I don't set options.Keys

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at JWT.Internal.EncodingHelper.GetBytes(String[] input)
   at JWT.JwtDecoderExtensions.DecodeToObject[T](IJwtDecoder decoder, String token, String[] keys, Boolean verify)
   at JWT.JwtAuthenticationHandler.GetAuthenticationResult(String header)

```

abatishchev:
@wtjones sorry! didn't realize the handlers calls an extension method first, then the decoder itself. I've updated once again, [it's being built](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=851&view=artifacts&pathAsName=false&type=publishedArtifacts) now.

wtjones:
Much better! As long as I give it an empty array, I no longer get the error.

Before:
`options.Keys = new string[] { " "};`
Now:
`options.Keys = new string[] {};`

wtjones (approved):
This allows for an empty Keys property in the options object. An empty array is still required at a minimum.

abatishchev:
@wtjones can you please post the exception stack trace where it complains about `null`?

wtjones:
@abatishchev the stack info you requested:
```
info: JWT.JwtAuthenticationHandler[2]
      Error decoding JWT: Object reference not set to an instance of an object., returning failure
System.NullReferenceException: Object reference not set to an instance of an object.
   at JWT.Internal.EncodingHelper.GetBytes(String[] input)
   at JWT.JwtDecoderExtensions.DecodeToObject[T](IJwtDecoder decoder, String token, String[] keys, Boolean verify)
   at JWT.JwtAuthenticationHandler.GetAuthenticationResult(String header)
```

abatishchev:
@wtjones I updated the tests to use RS256 algorithm and found where's the issue. A fix is being built, please [its drop](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=853&view=artifacts&pathAsName=false&type=publishedArtifacts) once complete.

abatishchev:
Perfect! Thanks for the confirmation, @wtjones.

Issue #358, opened 2021-11-10
Add algorithm "none" and generate custom header


hartmark:
We have some middleware where we use "unsecured jwts"

According to spec it's a valid alg:
https://datatracker.ietf.org/doc/html/rfc7519#section-6

A method named .WithoutAlgorithm() and making .Secret optional it it set would work for us.

We also need to add some custom values into the header.

We use the following code now as workaround:

```c#
        public string JwtToken
        {
            get
            {
                var token = JwtBuilder.Create()
                    .WithAlgorithm(new HMACSHA256Algorithm())
                    .WithSecret(Secret);
                AddHeaderClaims(token);
                AddPayloadClaims(token, Payload);
                var tokenWithValidPayload = token.Encode();
                
                var token2 = JwtBuilder.Create()
                        .WithAlgorithm(new HMACSHA256Algorithm())
                        .WithSecret(Secret);
                    
                AddHeaderClaims(token2);
                AddPayloadClaims(token2, Header);
                var tokenWithHeaderAsPayload = token2.Encode();

                var headerPart = tokenWithHeaderAsPayload.Split('.')[1];
                var payloadPart = tokenWithValidPayload.Split('.')[1];
                var signature = string.Empty;
                return $"{headerPart}.{payloadPart}.{signature}";
            }
        }
```

PR #360, opened 2021-11-17
Added support to specify custom headers using fluent calls #358

hartmark:


src/JWT/Algorithms/NoneAlgorithm.cs
```diff
@@ -10,7 +10,7 @@ namespace JWT.Algorithms
     public class NoneAlgorithm : IJwtAlgorithm
     {
         /// <inheritdoc />
-        public string Name => nameof(JwtAlgorithmName.None);
+        public string Name => "none";
```
hartmark, line 13: According to the RFC they type "none" in all lowercase so I change it here too.

abatishchev (approved):

abatishchev:
Pushed to NuGet as [8.5.1](https://www.nuget.org/packages/JWT/8.5.1)

PR #362, opened 2021-11-29
Added RSA algorithm for 1024, 2048, 4096 key lengths

Quentin-Harnay:
Adds implementation of the RSA algorithm with 1024, 2048 and 4096 lengths for the future.


abatishchev:
@Quentin-Harnay, awesome! thank you for the valuable contribution. I'll push it to NuGet.org shortly.

abatishchev (approved):

PR #370, opened 2022-01-21
Made JwtValidator.GetException() public

### LINKED ISSUE #369: How to get the TokenExpiredException on a token, without the signature verification
First of all thanks for the great library. I have an application which calls cyclically some web service api, secured with a JWT token. I would like to check if the token has expired before each call to the web service. For this I would like to catch the TokenExpiredException when decoding the jwt token. The problem is that the "secret" of the jwt token is not known on the application side, so that I am getting the SignatureVerificationException, even for the expired tokens.
Is there any workaround possible to get the TokenExpiredException without having the secret?

abatishchev:
Resolves #369.

### DIFF
--- a/src/JWT/IBase64UrlEncoder.cs
+++ b/src/JWT/IBase64UrlEncoder.cs
@@ -6,12 +6,12 @@ namespace JWT
     public interface IBase64UrlEncoder
     {
         /// <summary>
-        /// Encodes the byte array to a Base64 string.
+        /// Encodes the byte array to a base64 string.
         /// </summary>
         string Encode(byte[] input);
 
         /// <summary>
-        /// Decodes the Base64 string to a byte array.
+        /// Decodes the base64 string to a byte array.
         /// </summary>
         byte[] Decode(string input);
     }
--- a/src/JWT/JWT.csproj
+++ b/src/JWT/JWT.csproj
@@ -27,7 +27,7 @@
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json;authorization</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>8.8.1</Version>
+    <Version>8.8.2</Version>
     <FileVersion>8.0.0.0</FileVersion>
     <AssemblyVersion>8.0.0.0</AssemblyVersion>
     <RootNamespace>JWT</RootNamespace>
--- a/src/JWT/JWT.csproj
+++ b/src/JWT/JWT.csproj
@@ -27,7 +27,7 @@
     <Authors>Alexander Batishchev, John Sheehan, Michael Lehenbauer</Authors>
     <PackageTags>jwt;json;authorization</PackageTags>
     <PackageLicense>CC0-1.0</PackageLicense>
-    <Version>8.8.2</Version>
+    <Version>8.9.0</Version>
     <FileVersion>8.0.0.0</FileVersion>
     <AssemblyVersion>8.0.0.0</AssemblyVersion>
     <RootNamespace>JWT</RootNamespace>
--- a/src/JWT/JwtBase64UrlEncoder.cs
+++ b/src/JWT/JwtBase64UrlEncoder.cs
@@ -10,7 +10,7 @@ using static System.String;
 namespace JWT
 {
     /// <summary>
-    /// Base64 encoding/decoding implementation according to the JWT spec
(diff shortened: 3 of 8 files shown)


abatishchev:
@abrasat please take a look. You can consume a nuget package produces by the private build to confirm whether it works. If yes, I can check it and push to NuGet.

abrasat:
@abatishchev  Thanks for the update. You need to add the public method also into the interface declaration:
```
    public interface IJwtValidator
    {
.............
        bool TryValidate(string payloadJson, IAsymmetricAlgorithm alg, byte[] bytesToSign, byte[] decodedSignature, out Exception ex);

        Exception GetValidationException(string payloadJson);
    }
```

Another problem is the static helper method GetString() from EncodedHelper.cs, used at the payload decoding. The method is public, but the EncodingHelper static class is marked internal:
```
    internal static class EncodingHelper
    {
        private static readonly UTF8Encoding utf8Encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
....
        public static string GetString(byte[] bytes) =>
            utf8Encoding.GetString(bytes);
    }
```

It works only if the GetString() method from UTF8Encoding is used directly from the application code :

```
   IDateTimeProvider provider = new UtcDateTimeProvider();
   IJsonSerializer serializer = new JsonNetSerializer();
   IJwtValidator validator = new JwtValidator(serializer, provider);
   IBase64UrlEncoder urlEncoder = new JwtBase64UrlEncoder();

   try
   {
       var jwtParts = new JwtParts(token);
       UTF8Encoding utf8Encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
       var decodedPayload = utf8Encoding.GetString(urlEncoder.Decode(jwtParts.Payload));
       var ex = validator.GetValidationException(decodedPayload);
       if (ex is object)
          throw ex;
   }
   catch (TokenExpiredException te)
   {
        Console.WriteLine("Token has expired");
   }
```

abatishchev:
Let me reply one by one:

> You need to add the public method also into the interface declaration:

No, I don't want to expose it via the interface, it's a helper method, not the general API contract. Few other methods are exposed the same way already.
Moreover, updating the interface is a breaking change and requires major version bump.

> The method is public, but the EncodingHelper static class is marked internal.

It's a two-line helper that helps the library to avoid duplication, it must be easy to have it in your code.
I'd like to avoid exposing to much of internals just in the sake of exposing.

> It works only if the GetString() method from UTF8Encoding is used directly from the application code :
What I think I can do is to add an overload that would do it for you. Hang on.

abrasat:
@abatishchev Ok, I understand your point. No problem to call the method from JwtValidator instead of using the interface

abatishchev:
With the new overload your code would look like this:

```c#
byte[] bytes = urlEncoder.Decode(jwtParts.Payload)
Exception ex = validator.GetValidationException(bytes);
```

I can't encapsulate the whole thing because `JwtValidator` doesn't hold a reference on `IBase64UrlEncoder`.

But you can do something like this:

```c#
class NoSignatureJwtValidator // can implement the interface to look like a decorator (but not really be one)
{
    private readonly JwtValidator _innerValidator; // unfortunately has to be the concrete class
    private readonly IBase64UrlEncoder _urlEncoder;

    public NoSignatureJwtValidator(JwtValidator innerValidator, IBase64UrlEncoder urlEncoder)
    {
        _innerValidator = innerValidator;
        _urlEncoder = urlEncoder;
    }

        public Exception GetValidationException(JwtParts parts)
        {
            byte[] bytes = _urlEncoder.Decode(jwtParts.Payload)
            return base.GetValidationException(bytes);
        }

}
```

abatishchev:
Hmm, on other hand I might be able to add the dependency since it's an implementation, not the interface itself.

abrasat:
@abatishchev Both possibilities (1 line of code + helper class vs 2 lines of code) look good to me. Thanks!

abatishchev:
Pushed an update. Check this out.

abatishchev:
Here are the [latest build artifcats](https://abatishchev.visualstudio.com/OpenSource/_build/results?buildId=938&view=artifacts&pathAsName=false&type=publishedArtifacts), please try them out.

abrasat:
@abatishchev I cannot check the build, but the new code looks good (with the TimeMargin fix).

abatishchev:
@abrasat I'll push to Nuget.org shortly.

But I forget to add some tests. Can you please contribute and add some? Thanks!

PR #376, opened 2022-02-22
Added token validation parameters

abatishchev:


abatishchev:
Release to nuget.org as [JWT 9.0.0-beta2](https://www.nuget.org/packages/JWT/9.0.0-beta2)

PR #400, opened 2022-06-04
Harre/null check 1

hartmark:
Added some tests for none algorithm and also implemented that signature MUST not be sent according to spec in:
https://datatracker.ietf.org/doc/html/rfc7519#section-6
https://datatracker.ietf.org/doc/html/rfc7518#section-3.6


hartmark:
Why was this closed?

hartmark:
I see that the original branch that your had was merged into main and was removed, so perhaps this is automatically closed as well.

Should I create a new PR to main instead if these changes are welcome?

abatishchev:
I didn't close it, must be automatically. Yes, please open a new PR. Sorry for the churn, all changes are welcome!

PR #401, opened 2022-06-04
Replaced NewtonSoft.Json with System.Text.Json

### LINKED ISSUE #364: Replace Newtonsoft with System.Text.Json
Hi!

Please update to .net 6 and replace Newtonsoft with System.Text.Json.

Thanks.

hartmark:
Resolves #364.

### DIFF
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -1,9 +1,8 @@
 ﻿using System;
 using System.Linq;
 using JWT.Algorithms;
-using JWT.Serializers;
-
 using static JWT.Internal.EncodingHelper;
+using static JWT.Serializers.JsonSerializerFactory;
 
 namespace JWT.Builder
 {
@@ -18,7 +17,7 @@ namespace JWT.Builder
         private IJwtDecoder _decoder;
         private IJwtValidator _validator;
 
-        private IJsonSerializer _serializer = new JsonNetSerializer();
+        private IJsonSerializer _serializer = CreateSerializer();
         private IBase64UrlEncoder _urlEncoder = new JwtBase64UrlEncoder();
         private IDateTimeProvider _dateTimeProvider = new UtcDateTimeProvider();
         private ValidationParameters _valParams = ValidationParameters.Default;
@@ -73,9 +72,6 @@ namespace JWT.Builder
         /// <summary>
         /// Sets JWT serializer.
         /// </summary>
-        /// <remarks>
-        /// If not set then default <see cref="JsonNetSerializer" /> will be used.
-        /// </remarks>
         /// <returns>Current builder instance</returns>
         public JwtBuilder WithSerializer(IJsonSerializer serializer)
         {
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
 
@@ -25,7 +26,7 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
+  <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
@@ -35,4 +36,5 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
   </ItemGroup>
+
 </Project>
--- a/JWT.sln
+++ b/JWT.sln
@@ -49,6 +49,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{44284230
 		tests\Directory.Build.props = tests\Directory.Build.props
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{C86A941F-F655-4EF5-A6A1-47BD4A7FC748}"
+	ProjectSection(SolutionItems) = preProject
+		src\Directory.Build.targets = src\Directory.Build.targets
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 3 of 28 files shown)


abatishchev:
Yay, it's working! At least it compiles :) But now tests fail, and looks like because of the differences between the two serializers.

Here's the list of failed tests: https://dev.azure.com/abatishchev/OpenSource/_build/results?buildId=1081&view=ms.vss-test-web.build-test-results-tab

hartmark:
I added the missing serializer/deserializer for Dictionary<string, object> so it gets the same behaviour as Json.net.

It passes the tests* now here on my arch linux using rider.

*= 19 tests fail on Linux, but they did that before this change too. I'll investigate further when I get the time.

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -296,6 +297,17 @@ public T Decode<T>(string token)
             return _decoder.DecodeToObject<T>(token, _secrets, _valParams.ValidateSignature);
         }
 
+        private static IJsonSerializer CreateSerializer()
```
abatishchev, line 300: Should we move this factory method to `JsonSerializerFactory`? Right now it's present in every class that uses `IJsonSerializer`.

hartmark: Yeah, that would be nice. I'm going to bed now, pushing a little fix for unittest on linux, hopefully it will work on windows too :)

abatishchev:
Can you please rebase the branch onto the latest main?

hartmark:
I have rebased now, can you double-check if I have missed something. You can squash the commits after that so we have a cleaner change-log if we need to rebase from main again.

abatishchev:
Build failed with a cryptic error:

> ##[error]C:\hostedtoolcache\windows\dotnet\sdk\6.0.300\Microsoft.Common.CurrentVersion.targets(1221,5): Error MSB3644: The reference assemblies for .NETFramework,Version=v3.5 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks

Since it's a branch in your fork, it's bit tricky to clone it locally. But I should be able to do that, will give it a try.

abatishchev:
I fixed the build an now only tests failing are in AspNetCore extensions project. Does it need any additional configuration?

abatishchev:
Fixed the tests, had to make a breaking change in AspNetCore extensions. Need to update tests to test both serialziers to make sure we don't break either.

hartmark:
Great find about the breaking test. I think this is a good change however as this allows to have this kind of claim.

```
{
 "fooClaim": { "property1" : "value", "property2" : "value2" ...
}
```

hartmark:
I have added a method to IJsonSerializer to instruct if the payload should be serialized as camelCase or PascalCase, I hadded some tests for it too.

I however noticed that the old test projects didn't trigger on the NotImplementedException in JsonNetSerializer.

I saw that the test projects had all set net462 as their TargetFramework, changing it to a version lover than net462 triggered the exception for JsonNetSerializer.

What's the ambition of supporting old .net frameworks?
According to Microsoft everything below .net fw 4.6.2 is EOL.
https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework

I think that as long as it still works on old .net frameworks it doesn't hurt that this supports them.
But, on the flip side. If we just support .net fw 4.6.2 we can just drop Newtonsoft completely as System.Text.Json supports .net fw 4.6.2 as lowest.

hartmark:
I have added another test for encoding using fluent call, It seems to fail encoding properly.

I need to go to bed now so I'll push and leave it up here if you find anything. There's end of school term tomorrow so I don't know how much time I have to investigate further this weekend.

Plus we should perhaps add the license for the SystemTextJsonDictionaryStringObjectJsonConverter class I added:
https://github.com/joseftw/JOS.SystemTextJsonDictionaryStringObjectJsonConverter/blob/develop/LICENSE

abatishchev:
> I have added a method to IJsonSerializer to instruct if the payload should be serialized as camelCase or PascalCase, I hadded some tests for it too.

Let's don't expose such property on the interface, it's too implementation specific, we should be able to set it somehow else. But main question - can you explain why it's needed?

> What's the ambition of supporting old .net frameworks? According to Microsoft everything below .net fw 4.6.2 is EOL.

Yes, I know it's EOL already. But the support was added and kept intentionally. Every time I attempted to drop older frameworks - users would raise an issue, ask to put it back. This library has a relatively-small user base so I'd like to cater to everyone who still uses it. Keeping .NET 3.5/4.0 around didn't bother me so far. Meanwhile it's ok to exclude them from newer features if necessary.

Overall I'd suggest to maybe split the PR as it grows in the amount and complexity of the changes.

abatishchev:
Here's the trick that sets the target framework to .NET 3.5/4.0:

```xml
 <ItemGroup>
    <ProjectReference Include="..\..\src\JWT\JWT.csproj">
      <SetTargetFramework>TargetFramework=net40</SetTargetFramework>
    </ProjectReference>
  </ItemGroup>
```

I verified yet another time recently, the resulting binaries do indeed reference the BCL of the said version.

hartmark:
> Here's the trick that sets the target framework to .NET 3.5/4.0:
> 
> ```
>  <ItemGroup>
>     <ProjectReference Include="..\..\src\JWT\JWT.csproj">
>       <SetTargetFramework>TargetFramework=net40</SetTargetFramework>
>     </ProjectReference>
>   </ItemGroup>
> ```
> 
> I verified yet another time recently, the resulting binaries do indeed reference the BCL of the said version.

Alright, but it still choses the System.Text.Json serializer as you can see at the failed test on .net40 test-project

hartmark:
I will try to come up with a solution for the serializer settings. We should perhaps use the ms dependency injection nuget as it seems the tests now fail that the settings have changed on a already used serializer.

If we use a DI component is easy to setup that it should use a new instance per usage

hartmark:
I can move the serializer settings to another PR so it's less code changed at once

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -251,6 +248,22 @@ public string Encode()
             return _encoder.Encode(_jwt.Header, _jwt.Payload, _secrets?[0]);
         }
 
+        public string Encode<T>(T payload)
+        {
+            EnsureCanEncode();
+
+            var payloadAsDictionary = payload.GetType()
```
abatishchev, line 255: Not sure if this will work. You basically doing a native serialization. What if the value is not a primitive type? I'd not add it, at least not in this PR.

abatishchev (changes requested):

hartmark:
Removed last commit

hartmark:
I have merged with latest from main as this branch had conflicts. I don't know the versioning rules so I just used same version as in main.

Should we squash the commits and try to get this merged into main?

abatishchev:
> I don't know the versioning rules so I just used same version as in main

Please bump to the next beta version. That would be `10.0.0-beta3`>

> Should we squash the commits

Up to you, I'll squash them on the PR completion

> and try to get this merged into main?

Sorry for making you waiting! I appreciate your contribution and the following up too. I don't forget, just didn't have time yet (and empty enough mind, I have a very cool but also a very demanding job, what keeps me preoccupies even more than I wish it would) to go through the chances for another, final review. Hopefully will have time by the end of this weekend.

Thanks again for the collaboration, I'll review and approve the chance as soon as I can.

hartmark:
No worries, I tend to keep myself occupied with other stuff too sometimes.

I'll squash the commits after work so we have more easily overview over the pr

hartmark:
I have squashed and bumped the version now, please check if I have missed anything

abatishchev (dismissed):

abatishchev:
Yay, awesome! Greatly appreciate your contribution, also your persistence and for bearing with me until we've made it happen.

Will push to NuGet shortly.

hartmark:
It was a fun collaboration and I guess I have been under-stimulated with coding at work so this was a nice challenge :)

PR #402, opened 2022-06-04
Added some checks for none algorithm

hartmark:
Added some tests for none algorithm and also implemented that signature MUST not be sent according to spec in:
- https://datatracker.ietf.org/doc/html/rfc7519#section-6
- https://datatracker.ietf.org/doc/html/rfc7518#section-3.6

### DIFF
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -212,20 +212,8 @@ namespace JWT.Builder
         /// Instructs whether to verify the JWT signature.
         /// </summary>
         /// <returns>Current builder instance</returns>
-        public JwtBuilder WithVerifySignature(bool verify)
-        {
-            if (_algorithm is not NoneAlgorithm)
-            {
-                _valParams = _valParams.With(p => p.ValidateSignature = verify);
-            }
-            else
-            {
-                if (verify)
-                    throw new InvalidOperationException("Verify signature is not allowed for algorithm none");
-            }
-
-            return this;
-        }
+        public JwtBuilder WithVerifySignature(bool verify) =>
+            WithValidationParameters(p => p.ValidateSignature = verify);
 
         /// <summary>
         /// Sets the JWT signature validation parameters.
@@ -235,7 +223,11 @@ namespace JWT.Builder
         /// <returns>Current builder instance</returns>
         public JwtBuilder WithValidationParameters(ValidationParameters valParams)
         {
+            if (valParams.ValidateSignature && _algorithm is NoneAlgorithm)
+                throw new InvalidOperationException("Verify signature is not allowed for algorithm None");
+
             _valParams = valParams;
+
             return this;
         }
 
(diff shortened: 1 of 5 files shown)


src/JWT/Builder/JwtBuilder.cs
```diff
@@ -159,6 +159,12 @@ public JwtBuilder WithAlgorithmFactory(IAlgorithmFactory algFactory)
         public JwtBuilder WithAlgorithm(IJwtAlgorithm algorithm)
         {
             _algorithm = algorithm;
+
+            if (String.Equals(algorithm?.Name, nameof(AlgorithmName.None), StringComparison.OrdinalIgnoreCase))
```
abatishchev, line 163: Do you know if this can be `algorithm is NoneAlgorithm`?

We can add a private ctor to it so there will only `NoneAlgorithm.Instance`.

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -192,8 +198,15 @@ public JwtBuilder WithSecret(params byte[][] secrets)
         /// Instructs to verify the JWT signature.
         /// </summary>
         /// <returns>Current builder instance</returns>
-        public JwtBuilder MustVerifySignature() =>
-            WithVerifySignature(true);
+        public JwtBuilder MustVerifySignature()
+        {
+            if (String.Equals(algorithm?.Name, nameof(AlgorithmName.None), StringComparison.OrdinalIgnoreCase))
```
abatishchev, line 203: Please move this check to `WithVerifySignature(bool)`

abatishchev (changes requested):

src/JWT/Builder/JwtBuilder.cs
hartmark: Changing to just _algorithm is NoneAlgorithm works well.

src/JWT/Builder/JwtBuilder.cs
hartmark: done in latest push

src/JWT/Builder/JwtBuilder.cs
hartmark: I have pushed the simplified check, the singleton Instance is left to do.

abatishchev: Singleton is up to you, if it's something you feel would be useful

abatishchev:
Awesome, thanks much for the contribution and for quickly making the suggested changes too!

Please bump the version slightly inside JWT.csproj and it's almost good to go.

src/JWT/Builder/JwtBuilder.cs
hartmark: I'll skip the singleton for now and will revisit it in future if needed

hartmark:
Cool, I'll look at bumping the version later tonight.

Btw, have your had any chance to look at my other PR for system.text.json #401

abatishchev (dismissed):

abatishchev:
> Btw, have your had any chance to look at my other PR for system.text.json #401 

Will very definitely do!

hartmark:
There, I'm done now :)

Reduced some duplicated code in JwtDecoder and reverted WithValidator to previous state

abatishchev (approved):

PR #415, opened 2022-06-18
Added Encode<T>(T payload) so we have symmetry with Decode<T>(string token)

hartmark:


src/JWT/Builder/JwtBuilder.cs
```diff
@@ -247,6 +249,22 @@ public string Encode()
             return _encoder.Encode(_jwt.Header, _jwt.Payload, _secrets?[0]);
         }
 
+        public string Encode<T>(T payload)
```
abatishchev, line 252: Should this be just a wrapper and the actual implementation be in JwtEncoder or rather IJwtEncoderExceptions? So it won't be mandatory to the builder, so far it was not.

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -247,6 +249,22 @@ public string Encode()
...
+        {
+            EnsureCanEncode();
+
+            var dic = payload.GetType()
+                             .GetProperties(BindingFlags.Instance | BindingFlags.Public)
+                             .ToDictionary(prop => prop.Name, prop => prop.GetValue(payload, null));
+
+            foreach (var pair in dic)
+            {
+                _jwt.Payload.Add(pair.Key, pair.Value);
```
abatishchev, line 262: Should we add method `AddRange()` to avoid this loop?

If not, please see if you can deconstruct to a value tuple `(key, value)`

src/JWT/Serializers/Converters/DictionaryStringObjectJsonConverter.cs
```diff
@@ -117,7 +119,15 @@ private static void HandleValue(Utf8JsonWriter writer, string key, object object
                 }
                 default:
                 {
-                    writer.WriteNullValue();
+                    var payloadPartAsDictionary = objectValue.GetType()
```
abatishchev, line 122: Apparently the code is duplicated. Let's create an internal helper method.

Or should we have this code here in the first place? Writing `null` seems to be a more expected behavior. Until we provide an option to control it

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -247,6 +249,22 @@ public string Encode()
             return _encoder.Encode(_jwt.Header, _jwt.Payload, _secrets?[0]);
         }
 
+        public string Encode<T>(T payload)
+        {
+            EnsureCanEncode();
+
+            var dic = payload.GetType()
+                             .GetProperties(BindingFlags.Instance | BindingFlags.Public)
```
abatishchev, line 257: I think it should be rather `typeof(T)` which would be a simple syntax sugar on top of `Encode(Type payloadType, object payload)` as it's already in `IJwtDecoderExtensions`, check it out.

abatishchev (changes requested):

src/JWT/Builder/JwtBuilder.cs
hartmark: I'm not sure, I guess it's good to have the option to use either version

src/JWT/Builder/JwtBuilder.cs
hartmark: I'll look into it

src/JWT/Serializers/Converters/DictionaryStringObjectJsonConverter.cs
hartmark: Yeah, that sounds like a more safe approach

src/JWT/Builder/JwtBuilder.cs
hartmark: Payload is a dictionary so it has no AddRange
https://stackoverflow.com/questions/6050633/why-doesnt-dictionary-have-addrange

src/JWT/Serializers/Converters/DictionaryStringObjectJsonConverter.cs
hartmark: Added a test to describe what I want to achieve. I want to be able to add a payload like this:
```
{
  "ExtraClaim": {
    "NestedProperty1": "Foo",
    "NestedProperty2": 3
  },
  "FirstName": "Jesus",
  "Age": 33
} 
```

src/JWT/Builder/JwtBuilder.cs
hartmark: I don't follow exactly how you want it

abatishchev: Made the change, check it out. Does it make sense to you?

src/JWT/Serializers/Converters/DictionaryStringObjectJsonConverter.cs
abatishchev: And the only way to achieve it is to have this code in the `default` case? It worries me, to be frank, as some sort of over-engineering.

Can you explain why do you need a complex object as a claim? Usually claims are a list of key-value pairs which are primitives or just strings.

src/JWT/Builder/JwtBuilder.cs
hartmark: Yeah, i understand now 😀

We should have a test for it to see what happens if you send Encode(type, payload) where they are different types.

src/JWT/Serializers/Converters/DictionaryStringObjectJsonConverter.cs
abatishchev: Sounds reasonable and worth giving it a try

abatishchev (dismissed):

src/JWT/Builder/JwtBuilder.cs
hartmark: added test

hartmark:
Added some tests, please check again if all is good.

abatishchev:
Pushed few polishing changes, otherwise looks like it's good to go. Yay!

abatishchev (approved):

abatishchev:
Released to NuGet as [10.0.0-beta5](https://www.nuget.org/packages/JWT/10.0.0-beta5)

PR #429, opened 2022-07-16
Add null checks for JwtValidatorTests

KurnakovMaksim:
Under this lines we already have null checks: `ex?.Message`
I think my changes make code more safely.


abatishchev:
Hi Maksim, thanks for the contribution!
`Should()` is an extension method coming from [FluentAssertions library](https://www.nuget.org/packages/FluentAssertions/). Extension methods invocation won't fail with a NRE if the call site is null.
The purpose of the method to fail with a meaningful error if it's null. However your change would make it a no-op if it is, defeating the purpose.
Sorry, I have to reject your PR :(
But I appreciate you going through the codebase and looking for potential issue! Please do more of that and I'm happy to review and accept purposed changes.

PR #433, opened 2022-08-25
Added option to select serializer

### LINKED ISSUE #420: Library should include both Json.Net and System.Text.Json serialziers
After the recent changes, it includes one or another.

@hartmark we've made it too opinionated: either one or another. How about somehow allow both options on the platform whether that's possible?

hartmark:
Fixes #420 

* Changed so we always import JSON.net and added conditional just for System.Text.Json on newer dotnets.
* Opened up the JsonSerializerFactory so we can set another serializer if wanted.

I wanted to create a simple test for checking that the correct class was used as default.
I however noted that all tests returned it was using SystemTextSerializer even on the older dotnets.
See the Serializer_Should_Use_Correct_Default test.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -4,14 +4,8 @@
   <Choose>
     <When Condition="'$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net6.0'">
       <PropertyGroup>
-        <DefineConstants>$(DefineConstants);SYSTEM_TEXT_JSON</DefineConstants>
+        <DefineConstants>$(DefineConstants);MODERN_DOTNET</DefineConstants>
       </PropertyGroup>
     </When>
-    <Otherwise>
-      <PropertyGroup>
-        <DefineConstants>$(DefineConstants);NEWTONSOFT_JSON</DefineConstants>
-      </PropertyGroup>
-    </Otherwise>
   </Choose>
-
 </Project>
--- a/src/JWT.Extensions.DependencyInjection/JWT.Extensions.DependencyInjection.csproj
+++ b/src/JWT.Extensions.DependencyInjection/JWT.Extensions.DependencyInjection.csproj
@@ -30,8 +30,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.25" />
-    <PackageReference Condition="$(DefineConstants.Contains(SYSTEM_TEXT_JSON))" Include="System.Text.Json" Version="6.0.4" />
-    <PackageReference Condition="$(DefineConstants.Contains(NEWTONSOFT_JSON))" Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Condition="$(DefineConstants.Contains(MODERN_DOTNET))" Include="System.Text.Json" Version="6.0.4" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -78,6 +78,9 @@ namespace JWT.Builder
         public JwtBuilder WithSerializer(IJsonSerializer serializer)
         {
             _serializer = serializer;
+#if MODERN_DOTNET
+            Serializer = serializer;
+#endif
             return this;
         }
 
--- a/src/JWT/Serializers/JsonSerializerFactory.cs
+++ b/src/JWT/Serializers/JsonSerializerFactory.cs
@@ -1,16 +1,28 @@
 ﻿namespace JWT.Serializers
 {
-    internal static class JsonSerializerFactory
+    public static class JsonSerializerFactory
     {
-        public static IJsonSerializer CreateSerializer()
+        private static IJsonSerializer _serializer;
+
+        public static IJsonSerializer Serializer
         {
-#if SYSTEM_TEXT_JSON
-            return new SystemTextSerializer();
-#elif NEWTONSOFT_JSON
-            return new JsonNetSerializer();
+            get
+            {
+#if MODERN_DOTNET
+                return _serializer ??= new SystemTextSerializer();
 #else
-            throw new System.NotSupportedException();
+                
+                return _serializer ??= new JsonNetSerializer();
+#endif
+            }
+#if MODERN_DOTNET
+            set
+            {
+                _serializer = value;
+            }
 #endif
         }
+
+        public static IJsonSerializer CreateSerializer() => Serializer;
     }
 }
--- a/tests/JWT.Tests.Common/SerializerTests.cs
+++ b/tests/JWT.Tests.Common/SerializerTests.cs
@@ -1,0 +1,17 @@
+﻿using System;
+using JWT.Serializers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace JWT.Tests
+{
+    [TestClass]
+    public class SerializerTests
+    {
+        [TestMethod]
+        public void Serializer_Should_Use_Correct_Default()
+        {
+            var serializer = JsonSerializerFactory.Serializer;
+            Console.WriteLine(serializer.GetType().Name);
+        }
+    }
+}
--- a/src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
+++ b/src/JWT.Extensions.AspNetCore/JWT.Extensions.AspNetCore.csproj
@@ -30,8 +30,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Authentication" Version="2.2.0" />
-    <PackageReference Condition="$(DefineConstants.Contains(SYSTEM_TEXT_JSON))" Include="System.Text.Json" Version="6.0.4" />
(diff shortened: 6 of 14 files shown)


src/JWT/Builder/JwtBuilder.cs
```diff
@@ -78,6 +78,9 @@ public JwtBuilder AddClaim(string name, object value)
         public JwtBuilder WithSerializer(IJsonSerializer serializer)
         {
             _serializer = serializer;
+#if MODERN_DOTNET
```
abatishchev, line 81: What's the point? Both a field and a property? Should we have just one?

src/JWT.Extensions.DependencyInjection/JWT.Extensions.DependencyInjection.csproj
```diff
@@ -30,8 +30,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.25" />
-    <PackageReference Condition="$(DefineConstants.Contains(SYSTEM_TEXT_JSON))" Include="System.Text.Json" Version="6.0.4" />
-    <PackageReference Condition="$(DefineConstants.Contains(NEWTONSOFT_JSON))" Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Condition="$(DefineConstants.Contains(MODERN_DOTNET))" Include="System.Text.Json" Version="6.0.4" />
```
abatishchev, line 33: Please bump version in all projects

src/JWT/Serializers/JsonSerializerFactory.cs
```diff
@@ -1,16 +1,28 @@
 ﻿namespace JWT.Serializers
 {
-    internal static class JsonSerializerFactory
+    public static class JsonSerializerFactory
     {
-        public static IJsonSerializer CreateSerializer()
+        private static IJsonSerializer _serializer;
+
+        public static IJsonSerializer Serializer
```
abatishchev, line 7: Would something like this better?

```csharp
public static IJsonSerializer Serializer { get; set; } = new ...
```

or

```csharp
public static IJsonSerializer Serializer => new ...
```

(I'm always confused between these two syntax options)

tests/JWT.Tests.Common/SerializerTests.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace JWT.Tests
+{
+    [TestClass]
+    public class SerializerTests
+    {
+        [TestMethod]
+        public void Serializer_Should_Use_Correct_Default()
+        {
+            var serializer = JsonSerializerFactory.Serializer;
+            Console.WriteLine(serializer.GetType().Name);
```
abatishchev, line 14: Please assert something. Or that's the test which always returns the same serializer?

src/JWT/Serializers/JsonSerializerFactory.cs
```diff
@@ -1,16 +1,28 @@
 ﻿namespace JWT.Serializers
 {
-    internal static class JsonSerializerFactory
+    public static class JsonSerializerFactory
```
abatishchev, line 3: Is it necessary to make it public?

src/JWT/Serializers/JsonSerializerFactory.cs
abatishchev: Or better let's make it a method, a factory with property looks more like a nonsense :)

abatishchev (changes requested):

src/JWT/Serializers/JsonSerializerFactory.cs
```diff
@@ -1,16 +1,28 @@
...
+        private static IJsonSerializer _serializer;
+
+        public static IJsonSerializer Serializer
         {
-#if SYSTEM_TEXT_JSON
-            return new SystemTextSerializer();
-#elif NEWTONSOFT_JSON
-            return new JsonNetSerializer();
+            get
+            {
```
abatishchev, line 10: I know why it always returns the same serializer. Because of this line, it's the compile-time condition. Tests that target .NET 4.0 are compiled against higher version but references lower in the runtime.

What we should do instead is to:
1. Provide a parameter in the ctor, e.g. `modernSerializer`
and/or
2. Provide two factories, each would always return its own serializer

abatishchev (changes requested):

src/JWT/Serializers/JsonSerializerFactory.cs
hartmark: Yeah, a method looks better in a factory.

src/JWT/Serializers/JsonSerializerFactory.cs
hartmark: It's used by JWT.Extensions.DependencyInjection to get the default serializer

src/JWT/Serializers/JsonSerializerFactory.cs
hartmark: I don't follow, I thought the compile constant was set to either value depending if it's below or above net462.

abatishchev: I mean the idea is to ship below net462 with only json.net, and above with both. Correct?

tests/JWT.Tests.Common/SerializerTests.cs
hartmark: I have updated the test to test against the identified running dotnet version. As you can see all .net fw tests uses same version:
Dotnet version: .NETFramework,Version=v4.6.2

image

I wanted to have a test where it was identified to be running .net46 for example and in that case return JsonNetSerializer as the serializer.

src/JWT/Serializers/JsonSerializerFactory.cs
hartmark: Yeah, that's the idea so that's why it just always imports the json.net nuget and an conditional for the systemtext.json

abatishchev: I think because of the combination of these two:

https://github.com/jwt-dotnet/jwt/blob/b38709443e6b96e5757f34b0f91beab869e03bcb/tests/JWT.Tests.Net40/JWT.Tests.Net40.csproj#L4

https://github.com/jwt-dotnet/jwt/blob/b38709443e6b96e5757f34b0f91beab869e03bcb/tests/JWT.Tests.Net40/JWT.Tests.Net40.csproj#L9

So in the build time it targets `net462` and in the runtime - `net40`.

That was a sort of a hack made by a contributor to make very old versions of .NET compilable in the modern environment.

hartmark: I got the test working with net461 so it did detect JSON.net as its default serializer. If I lowered the dotnet version any lower I got errors that other nuggets required a higher dotnet version.

hartmark:
It was indeed using the JSON.net serializer now with latest push in the net40 unittest project.

src/Directory.Build.targets
```diff
@@ -4,14 +4,8 @@
   <Choose>
     <When Condition="'$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net6.0'">
       <PropertyGroup>
-        <DefineConstants>$(DefineConstants);SYSTEM_TEXT_JSON</DefineConstants>
+        <DefineConstants>$(DefineConstants);MODERN_DOTNET</DefineConstants>
```
abatishchev, line 7: I wonder if there is a predefined constant we could use. No, nothing?

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -78,6 +78,9 @@ public JwtBuilder AddClaim(string name, object value)
         public JwtBuilder WithSerializer(IJsonSerializer serializer)
         {
             _serializer = serializer;
+#if MODERN_DOTNET
+            SetSerializer(serializer);
```
abatishchev, line 82: So this one calls the factory by the imported static method? Far from being obvious. Should we call the factory explicitly?

abatishchev: Another, bigger problem with this approach is that an instance methods changes the static state. Should it instead instantiate a factory with the given serializer and pass it whenever applicable?

src/JWT/Serializers/JsonSerializerFactory.cs
```diff
@@ -1,15 +1,22 @@
 ﻿namespace JWT.Serializers
 {
-    internal static class JsonSerializerFactory
+    public static class JsonSerializerFactory
     {
+        private static IJsonSerializer _jsonSerializer;
+
+#if MODERN_DOTNET
+        public static void SetSerializer(IJsonSerializer jsonSerializer)
+        {
```
abatishchev, line 9: I think it should be a ctor, used by the builder and anyone else.

src/Directory.Build.targets
hartmark: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives
We could use NET462_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NET6_0_OR_GREATER
to replicate same behavior and use that throughout the solution instead.

abatishchev: Nah, I think it's too long. I think i won't bring as many benefits, but will definitely hurt the readability.

We might be able to define our own `NET462_OR_GREATER` but then it would introduce local behavior, inconsistent with other OSS repos.

src/JWT/Serializers/JsonSerializerFactory.cs
hartmark: I made the JsonSerializerFactory non-static and it looks better now.

abatishchev:
Sorry to make you waiting! Will come back to it tonight, if not then shortly.

abatishchev (approved):

PR #439, opened 2022-09-22
Fixed handling IEnumerable by DictionaryStringObjectJsonConverter

hartmark:
Resolves #438.

PR #441, opened 2022-09-26
DictionaryStringObjectJsonConverter is only needed when deserializing

hartmark:
I posted about our finding of the bug #439 in the repo from where I found the code for handling Dictionary<string, object> serialization.

https://github.com/joseftw/JOS.STJ.DictionaryStringObjectJsonConverter/issues/2

Apparently the converter is only needed when reading serialized data so we can cleanup our code a bit.


abatishchev:
@hartmark can you please explain this change little further?

@MaxLevs would you be able to test it once released?

hartmark:
The guy that wrote the code pointed out that it's working for serialization but it's not recommended as it's slower.
See the description on this page:
https://github.com/joseftw/JOS.STJ.DictionaryStringObjectJsonConverter

abatishchev (approved):

PR #442, opened 2022-09-28
Use JOS.SystemTextJson.DictionaryStringObject.JsonConverter nuget

hartmark:
The code we use to convert Dictionary<string, object> is now contained in a nuget.
https://www.nuget.org/packages/JOS.SystemTextJson.DictionaryStringObject.JsonConverter
https://github.com/joseftw/JOS.SystemTextJson.DictionaryStringObject.JsonConverter


abatishchev:
To be frank, I'm hesitant to take a dependency on an external package. Currently the library is mostly self-contained, for everything other than Json.Net before and also System.Text.Json now. Plus it's just one class we need from it, I'd rather keep a copy of it.

hartmark:
The nuget is just that class, so nothing else is added we don't need. One advantage of using a nuget is that the code is clearly separated from this project and any bugs are fixed over there.

The version I copied from the repo has already diverged with the latest.

abatishchev:
Is this a source code and build-time package, or a runtime dependency which will be consumed by everyone using the library? I totally support the former and disagree with the latter.
We can reach out to the authors and ask to publish a resource code package.

hartmark:
It seems to be a regular nugget. I have no experience with resource packages but that sounds like what we would want.

Can you post an issue on his repo and request that? He seems to be quite fast to respond.

abatishchev:
Will do.

Issue #451, opened 2023-01-09
Support for .NET 7.0


PMendesWebDev:
Will there be a release for .NET 7.0?

abatishchev:
Hi and thanks for checking on.
While I'd agree that the new major version release could've been the best time to add a new target framework, my question and argument against doing that would be this: would it be really necessary and beneficial? In other words, what would be the difference for an application targeting 7.0 when the library targets 6.0 or 7.0? To my knowledge, it'll work the same way. No?

PMendesWebDev:
Hey there! 

Thanks for the fast reply.

I could be being a complete noob on this, but I'm not able to install JWT on my project because it doesn't target .NET 7.0 (at least it won't show in nuget library)

PMendesWebDev:
I used vs' nuget package console and it did let me install it.

I'll try to work with it and see if it's all good :)

Thanks for the reply!

abatishchev:
Happy to help, let me know if you run into any other issues.

PR #463, opened 2023-01-24
Added unit tests for .NET 7

hartmark:
Related to #451

### DIFF
--- a/.pipelines/build.yml
+++ b/.pipelines/build.yml
@@ -23,10 +23,17 @@ steps:
     installationPath: '$(Agent.ToolsDirectory)/dotnet'
 
 - task: UseDotNet@2
-  displayName: 'Install .NET 6.0.x'
+  displayName: 'Install .NET 6.x'
   inputs:
     packageType: sdk
-    version: '6.0.x'
+    version: '6.x'
+    installationPath: '$(Agent.ToolsDirectory)/dotnet'
+
+- task: UseDotNet@2
+  displayName: 'Install .NET 7.x'
+  inputs:
+    packageType: sdk
+    version: '7.x'
     installationPath: '$(Agent.ToolsDirectory)/dotnet'
 
 - task: NuGetToolInstaller@1
--- a/JWT.sln
+++ b/JWT.sln
@@ -55,6 +55,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{C86A941F-F65
 		src\Directory.Build.targets = src\Directory.Build.targets
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JWT.Tests.Net70", "tests\JWT.Tests.Net70\JWT.Tests.Net70.csproj", "{D7F24AC9-D178-4BAB-BF93-4BAD8028416D}"
+EndProject
 Global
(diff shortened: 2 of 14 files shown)


src/Directory.Build.targets
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Choose>
-    <When Condition="'$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net6.0'">
+    <When Condition="'$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net6.0' OR '$(TargetFramework)' == 'net7.0'">
```
abatishchev, line 5: I think some `|| NET7.0` are missing

tests/JWT.Tests.Common/JWT.Tests.Common.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net7.0;net6.0;netstandard2.0;net462</TargetFrameworks>
```
abatishchev, line 4: My main question would be why to do this? In other words, what's the benefit of targeting net7 if we don't call any net7 specific API?

abatishchev (changes requested):

src/Directory.Build.targets
abatishchev: Also please update the Extension projects

abatishchev (changes requested):

tests/JWT.Tests.Common/JWT.Tests.Common.csproj
hartmark: .net6 shouldn't have been needed too, as .netstandard2 should be enough. I'll check it out can be simplified

abatishchev: The .NET Standard is dead, the recommendation by the .NET team is fo target certain "full" .NET version, such as 3.1 or 6. And that was my plan going forward - to add the LTS versions, i.e. skip 7.0 until really needed, snd later add 8.0

hartmark: Alright, fair enough. I'll try remove .net7

hartmark:
@abatishchev From your comment that the nugget should only target LTS versions of .net I have changed this PR to just add the Test-project of .net7 so we can be certain that everything still works for anyone using this nuget on .net7

abatishchev (dismissed):

abatishchev:
@hartmark can you please see why on earth the packages restore fails? :(

abatishchev (approved):

hartmark:
Cool, nice cooperation 😎

Regarding nugget failure,  Microsoft has some issues earlier today, perhaps related. Good it's working now.

abatishchev:
Cool, thanks for the contribution, appreciate it as always.

Not sure, if it was a network issue or just missing additional framework installed by the build pipeline.

Issue #456, opened 2023-01-12
Decode<JwtHeader> not working


thoron:
### Regression issue in Decode
When upgrading from 9.0.3 to 10.0.0 decode header will cease to function correctly, all fields will be `null`:

```csharp
var tokenString = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
var serializer = new JsonNetSerializer();
var urlEncoder = new JwtBase64UrlEncoder();
var decoder = new JwtDecoder(serializer, urlEncoder);

var header = decoder.DecodeHeader<JwtHeader>(tokenString);
Console.WriteLine(header.Algorithm);
```
For 9.0.3
```
HS256
```
For 10.0.0
```
 // Empty string or null
```

abatishchev:
hi @thoron, what's your target platform?

I'm unable to reproduce the issue. I've put your token into [TestData](https://github.com/jwt-dotnet/jwt/blob/main/tests/JWT.Tests.Common/Models/TestData.cs#L44) and these Header [related tests](https://github.com/jwt-dotnet/jwt/blob/main/tests/JWT.Tests.Common/JwtDecoderTests.cs#L21). After updating the asserts accordingly they all have passed.

thoron:
@abatishchev I encountered the issue when using a win11 computer. Now using a Linux computer (Pop 22.04) I cannot reproduce it either. Both were targeting net7.0.
I do not currently have access to a Windows computer, I will try to repro again on Monday and if I cannot not I will close this issue and blame cosmic background radiation.

abatishchev:
I have only Windows machine but I'll try to reproduce it on different frameworks too.

abatishchev:
I see, so you can reproduce it now.
Does it still happen if you downgrade the target version to net6.0? And please confirm it does not happen if you downgrade the version of JWT to 9.x.
It might be a bug in STJ, not JWT though :(

thoron:
Using .NET 7 SDK with target net6.0 can to reproduce the issue when using JWT 10.0.0. Downgrading fixes the issue.

PeterHagen:
[...]

PR #464, opened 2023-01-24
Disallow Encode(payload) with AddClaim(s)

hartmark:
Fixes for https://github.com/jwt-dotnet/jwt/issues/456.

### DIFF
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -269,27 +269,20 @@ namespace JWT.Builder
             return _encoder.Encode(_jwt.Header, _jwt.Payload, _secrets?[0]);
         }
 
-        public string Encode<T>(T payload) =>
-            Encode(typeof(T), payload);
-        
-        public string Encode(Type payloadType, object payload)
+        public string Encode(object payload)
         {
-            if (payloadType is null)
-                throw new ArgumentNullException(nameof(payloadType));
             if (payload is null)
                 throw new ArgumentNullException(nameof(payload));
-            
+
             EnsureCanEncode();
 
-            var dic = payloadType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                                 .ToDictionary(prop => GetPropName(prop), prop => prop.GetValue(payload, null));
-
-            foreach (var pair in dic)
+            if (_jwt.Payload.Any())
             {
-                _jwt.Payload.Add(pair.Key, pair.Value);
+                throw new NotSupportedException(
+                    "Having JWT as a key-value store and implicit payload is not supported.");
             }
 
-            return Encode();
+            return _encoder.Encode(_jwt.Header, payload, _secrets?[0]);
         }
 
         /// <summary>
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -282,7 +282,7 @@ namespace JWT.Builder
             EnsureCanEncode();
 
             var dic = payloadType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                                 .ToDictionary(info => GetPropName(info), prop => prop.GetValue(payload, null));
+                                 .ToDictionary(prop => GetPropName(prop), prop => prop.GetValue(payload, null));
 
             foreach (var pair in dic)
             {
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -504,7 +504,7 @@ namespace JWT.Builder
 #endif
                     default:
                     {
-                        throw new NotSupportedException($"{jsonSerializer.GetType()} is not supported");
+                        return prop.Name;
                     }
                 }
             }
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -282,7 +282,7 @@ namespace JWT.Builder
             EnsureCanEncode();
 
             var dic = payloadType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                                 .ToDictionary(info => GetPropName(info, _jsonSerializerFactory), prop => prop.GetValue(payload, null));
+                                 .ToDictionary(info => GetPropName(info), prop => prop.GetValue(payload, null));
 
             foreach (var pair in dic)
             {
--- a/src/JWT/Builder/JwtBuilder.cs
+++ b/src/JWT/Builder/JwtBuilder.cs
@@ -326,6 +326,14 @@ namespace JWT.Builder
             return _decoder.DecodeHeader<T>(token);
         }
 
+        public object Decode(string token, Type type)
+        {
+            EnsureCanDecode();
+
+            return _decoder.DecodeToObject(type, token, _secrets, _valParams.ValidateSignature);
+        }
+        
+        
         /// <summary>
         /// Decodes a token using the supplied dependencies.
         /// </summary>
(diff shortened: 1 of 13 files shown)


hartmark:
Fix for #456

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -468,5 +474,42 @@ private bool CanDecodeHeader()
...
+                        if (attribute is JsonPropertyNameAttribute stjProperty)
+                        {
+                            return stjProperty.Name;
+                        }
+                        break;
+                    }
+#endif
+                    default:
+                    {
+                        throw new NotSupportedException($"{jsonSerializer.GetType()} is not supported");
```
abatishchev, line 507: I don't think we should throw an exception here. It defeat the purpose of having an interface, we don't know if the customers have their own implementation, the idea is that they might.

abatishchev: Can the `default` case be just omitted so the method would return `prop.Name`?

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -276,7 +282,7 @@ public string Encode(Type payloadType, object payload)
             EnsureCanEncode();
 
             var dic = payloadType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
```
abatishchev, line 284: Overall, I think this is not the rich approach, the library shouldn't try to reinvent the wheel, must somehow let the serializer itself to deal with the type of the payload.

abatishchev (changes requested):

src/JWT/Builder/JwtBuilder.cs
abatishchev: @hartmark what you think about this variant of the method?

```c#
  public string Encode(Type payloadType, object payload)
  {
      if (payloadType is null)
          throw new ArgumentNullException(nameof(payloadType));
      if (payload is null)
          throw new ArgumentNullException(nameof(payload));
      
      EnsureCanEncode();

      return _encoder.Encode(_jwt.Header, payload, _secrets?[0]);
  }
```

which passes the newly added tests too!

abatishchev (changes requested):

src/JWT/Builder/JwtBuilder.cs
hartmark: good idea

src/JWT/Builder/JwtBuilder.cs
hartmark: The tests that just do .Encode without .AddClaim works

For example Encode_Should_Return_Token_With_Custom_Extra_Headers_Full_Payload_And_Claims fails as it adds the claim "ExtraClaim" first and that gets overwritten by .Encode method.

abatishchev (commented):
Ah, it ran into a merge conflict now :(

abatishchev:
Do I understand the origin of the issue correct is that it occurs only in the builder because its Encode() method "messed up" with the payload so a complex model would work?

If that's the case, what I think we should do: remove the code and move it to a custom serializer, which in turn just a decorator on the other one (maybe two decorators, one for the each underlying technology).

hartmark:
I'll check the conflict when I'm at a computer. Looks like something minor.

Yes, the issue is that Encode() with a parameter serializes the _jwt.Payload property.

While Encode() with a parameter serializes only what we pass to it, ignoring the _jwt.Payload if we don't either add what we pass to it or adds the claims manually.

hartmark:
I have reworked the Encode when also having added Claims so they are both serialized the same way as everything else.

I have for now just added another Encode in JwtEncoder that just accepts two payloads. And in that case just serializes them and uses the merge functionallity in JSON.net, for System.Text.Json it was a bit more tricky but after finding out you can use the JsonNode classes it was easier.

I think this is a more clean solution than iterating the payload with reflection and add all properties in the _jwt.Payload dictionary in JwtBuilder.

If this is an acceptable solution we could rework the JwtEncoder.Encode method to be less copy.paste

All tests pass also! :)

PS. For further reading about JsonNode and Merge
https://github.com/dotnet/runtime/issues/31433
https://makolyte.com/csharp-how-to-use-jsonnode-to-read-write-and-modify-json/

src/JWT/IJsonSerializer.cs
```diff
@@ -21,6 +21,14 @@ public interface IJsonSerializer
...
         /// <returns>Strongly-typed object.</returns>
         object Deserialize(Type type, string json);
+
+        /// <summary>
+        /// Merges two objects into a single JSON string.
+        /// </summary>
+        /// <param name="obj1">The first object to serialize</param>
+        /// <param name="obj2">The second object to serialize, note that any existing property in obj1 will be
+        /// <returns>JSON string</returns>
+        string MergeObjects(object obj1, object obj2);
```
abatishchev, line 31: I feel it's too hacky, in other words - leaky abstraction. Is there a way to support a complex model without adding this method?

src/JWT/Builder/JwtBuilder.cs
```diff
@@ -272,18 +278,16 @@ public string Encode(Type payloadType, object payload)
...
                 throw new ArgumentNullException(nameof(payload));
-            
-            EnsureCanEncode();
 
-            var dic = payloadType.GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                                 .ToDictionary(prop => prop.Name, prop => prop.GetValue(payload, null));
-
-            foreach (var pair in dic)
+            // Preserve old behaviour checking the type, is it needed?
+            if (payloadType != payload.GetType())
```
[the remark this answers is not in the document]
hartmark: I only added this because once test checked this. The previous implementation needed to know the type to be able to deconstruct the object with reflection.

The new implementation just uses generic serialisation so it's no longer needed to add the type.

src/JWT/IJsonSerializer.cs
hartmark: Fair enough, the solution before this commit also works so if that's a more acceptable solution we can revert back to that.

src/JWT/Builder/JwtBuilder.cs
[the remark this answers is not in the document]
abatishchev: I'd keep the library lightweight and relying on the underlying serialier to fail

src/JWT/IJsonSerializer.cs
abatishchev: Yes, please. Let ship a beta version and it works for everyone - then let's keep it simple.

However i cannot stress enough my appreciation for your contributions and trying to make the library better over and over again! hope you're ok with my ask to revert.

hartmark: Alright, no worries. It was an interesting solution but perhaps not the smallest code change 😀

I'll try get this sorted when I get the time, hopefully this evening.

hartmark: I have reverted that commit and rebased against latest on main now

src/JWT/Builder/JwtBuilder.cs
abatishchev, line 284: I still feel this is hacky. Is it possible to avoid doing it altogether?

hartmark: If we disallow using both .AddClaim()/.AddClaims() and ,Encode(payload) we should then be able to just send the payload directly to the serializer.

hartmark: Yes, that sounds like a reasonable approach

hartmark: I have added code to disallow mixing Encode(payload) with AddClaim/AddClaims

abatishchev (approved):

abatishchev:
Awesome, thank you!! Pushed some minor formatting changes. And dropped the beta prefix, I think it'll work as-is.

hartmark:
Cool, that was fast. I think we ended up with a cleaner solution with just sacrificing a minor nice-to-have functionallity

PR #467, opened 2023-03-01
Added note about breaking changes in 10.0.0

cmeeren:
Changing the default serializer is a breaking change and should be explicitly called out.

We were hit by outages in production because while I had carefully reviewed the changelog, this particular change was not mentioned, and therefore not tested. (Shame on us for not having a sufficiently comprehensive test suite to catch this, of course.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,12 @@
 
 # 10.0.0
 
-- **Breaking:** System.Text.Json is now the default serializer
+- **Breaking:** Made System.Text.Json the default serializer on the platforms where it's available
+- **Breaking:** Made verify=true by default in IJwtDecoder methods
+
 - Made NoneAlgorithm not requiring any keys as it is not signed
 - Added option to select default serializer, Newtonsoft.Json or System.Text.Json (#433)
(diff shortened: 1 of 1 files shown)


abatishchev:
An outage in production? Oh no! I'm sorry to hear/cause this, my apologies for not calling it out clearly. I'll update the changelog accordingly, thanks for the contribution!

For my better understanding, can you explain what was the scenario and how it got broken?

cmeeren:
> I'm sorry to hear/cause this

Again, not your fault. Shame on us for not having a sufficiently comprehensive test suite to catch this. If it were mentioned in the changelog, I would likely have checked a few more things manually and caught this, which is why I'm adding it. 🙂

> For my better understanding, can you explain what was the scenario and how it got broken?

I parsed the token to `Dictionary<string, obj>` and cast relevant payload items to `string`. Not sure why I went with that solution originally, but it worked, because apparently Newtonsoft.Json deserializes to a suitable underlying type when using `obj`. However, System.Text.Json deserialized to `JsonElement` or something like that, which made the code fail due to invalid cast.

PR #475, opened 2023-07-11
Added default ctor to JwtHeader and decorated it with [JsonConstructor]

### LINKED ISSUE #476: Dotnet 8.0 AOT Issues
Moving this into its own issue.

I was having AOT issues when trying to serialize the JwtHeader class. It is throwing an error for it missing a Parameterless constructor, as well as the JsonConstructor decorator. I described the issue here:

[Issue-474](https://github.com/jwt-dotnet/jwt/issues/474#issuecomment-1631476767)

This is the proposed PR to fix this particular issue.
[PR-475](https://github.com/jwt-dotnet/jwt/pull/475)

andreijw:
When running this project targetting dotnet 8.0 and building with AOT. If the JwtHeader class gets deserialized/serialized with the SystemTextSerializer.cs it throws an exception:

> Missing a Parameterless constructor <Class>".

I have added the System.Text.Json.JsonConstructor decorator with a fully qualified name since the Newtonsoft.Json namespace also have that same JsonConstructor decorator.  I also wrapped this behind the MODERN_DOTNET compiler directive to conserve backwards compatibility.

Resolves #476.


src/JWT/Builder/JwtHeader.cs
```diff
@@ -52,5 +52,10 @@ public class JwtHeader
         [JsonPropertyName("x5t")]
 #endif
         public string X5t { get; set; }
+
+#if MODERN_DOTNET
```
abatishchev, line 56: Please move it to the top of the class, so the ctor(s) would be before properties but after fields

abatishchev (changes requested):

abatishchev:
Thank you for the contribution! Please make additional necessary changes:

1. Bump the third element in the version inside JWT.csproj
2. Update the change log
3. Is it possible to add a unit test that would illustrate the issue? Why it doesn't fail on regular platforms?

andreijw:
@abatishchev  Ok will do! Might take a few days since I need to shift focus to some critical issues at work, but if I get the chance I will knock out these simple changes as requested. I appreciate the feedback

abatishchev:
Just did all of the above myself, but a unit test to demonstrate the issue. Please add. Thanks again for the contribution!

andreijw:
Hello @abatishchev I'm a little confused on the best way to add a unit test for this specific issue.

I only encounter this issue, when I have an application that I have compiled with AOT with some command similar to 
`dotnet publish SomeProject.csproj -c Release --runtime=linux-x64 /p:Version=VERSION` and it throws a runtime warning when the class (in this case the JWTHeader.cs) is serialized using System.Text.Json since it's missing a parameterless constructor and it's not annotated with the JsonConstructor attribute. 

Would adding a unit test that specifically invokes the parameterless constructor without any issues in both MODERN_DOTNET and non-MODERN_DOTNET platforms suffice your wants? I fear that to truly demonstrated the issue we need an integration / end-to-end test that has an AOT-compiled application and tries to serialize this class to throw a runtime error.

Please let me know your thoughts on this.

Regards

abatishchev:
> Would adding a unit test that specifically invokes the parameterless constructor without any issues in both MODERN_DOTNET and non-MODERN_DOTNET platforms suffice your wants?

Would be a good start! Something is better than nothing :)

> I fear that to truly demonstrated the issue we need an integration / end-to-end test that has an AOT-compiled application and tries to serialize this class to throw a runtime error.

I fear this too. But it's definitely more work and totally fine if won't be a part of this PR. Would be awesome if you still could do that later!

abatishchev (dismissed):

abatishchev (approved):

andreijw:
Hello @abatishchev

Is there anything special that we must do in order to publish the new nuget 10.0.3? I saw that you bumped the nuget version in the csrpoj file, but I still don't see the nuget released. I also checked the PR and merge build and ensured that they were both completed without any issues.

Regards,

Andrei Jaume

abatishchev:
Sorry, didn't have a chance to press the button yet. Will do it in a moment.

abatishchev:
[10.0.3](https://www.nuget.org/packages/JWT/10.0.3) should become available soon (once NuGet finishes indexing the package).

PR #495, opened 2024-04-04
Switch to Events Model to Allow IoC of Event Type

drusellers:
## Description

while working on some JWT code, I'd like to set the HttpContext.Items with a hydrated user. This follows a similar pattern seen in the Cookie and various other authentication systems.

We introduce a new `JwtAuthenticationEvents` type, that mirrors the exact same events. Following the Cookie Authentication I moved all of the `OnXxxx` to the `Events` object. I added [Obsolete] tags to the old `OnXxx` methods with instructions about what to use instead.

If we want we can also extend the `OAuthEvents` if that suits us as well.

I need to figure out how to best test this, but this is a pattern I've used to build my own Basic auth and MagicLink auth. Before I go any further I wanted to check in with the team on this direction.

The goal is to be able to inject my DbContext into the handler, so that I can populate the HttpContext.Items

```csharp
.AddJwt(JwtAuthenticationDefaults.AuthenticationScheme, options =>
    {
        options.VerifySignature = true;
        options.EventsType = typeof(MyJwtAuthenticationEvents);
    });
```

```csharp
public class MyJwtAuthenticationEvents : JwtAuthenticationEvents 
{
    readonly MyDbContext _context;

    public MyJwtAuthenticationEvents(MyDbContext context)
    {
        _context = context;
    }

    public override AuthenticateResult OnSuccessfulTicket(ILogger logger, AuthenticationTicket ticket)  
    {
        // do database stuff here
    }
}
```

## Checklist

- [ ] Tests added
- [ ] Version bumped
- [ ] Changelog updated


src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,99 @@
+namespace JWT.Extensions.AspNetCore;
```
abatishchev, line 1: not: for consistency with the rest of the codebase, please don't use implicit namespace

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        _logMissingHeader = LoggerMessage.Define(
+            LogLevel.Information,
+            10,
+            $"Header {nameof(HeaderNames.Authorization)} is empty, returning none");
+
+        _logIncorrectScheme = LoggerMessage.Define<string, string>(
+            LogLevel.Information,
+            11,
+            $"Header {nameof(HeaderNames.Authorization)} scheme is {{0}}, expected {{1}}, returning none");
```
abatishchev, line 26: The official recommendation is to use named placeholders, not numbered. Or this is the already existing (my) code? Should we change it?

src/JWT.Extensions.AspNetCore/JwtAuthenticationOptions.cs
```diff
@@ -8,6 +8,11 @@ namespace JWT.Extensions.AspNetCore
 {
     public class JwtAuthenticationOptions : AuthenticationSchemeOptions
     {
+        public JwtAuthenticationOptions()
```
abatishchev, line 11: I'd like initialization into the property rather than adding a ctor for the purpose of a single initialization

src/JWT.Extensions.AspNetCore/JwtAuthenticationOptions.cs
```diff
@@ -77,5 +87,15 @@ public class JwtAuthenticationOptions : AuthenticationSchemeOptions
         /// The default value is <see cref="Dictionary{String, String}" />.
         /// </remarks>
         public Type PayloadType { get; set; } = typeof(Dictionary<string, object>);
+
+
+        /// <summary>
+        /// Custom Event Overrides
+        /// </summary>
+        public new JwtAuthenticationEvents Events
```
abatishchev, line 95: Why `new` is needed?

abatishchev (changes requested):

abatishchev:
Hi Dru, thank you for the contribution! Let me understand the rationale a bit better, I'll come back with questions.

abatishchev:
Why `options.EventsType` is of type `Type`? How it's used later?

Can you post your old and new code side-by-side. I'm all for a better usage experience, so want to see understand how does the proposed change improve it.

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
drusellers: I think I just copied what was already there. But happy to upgrade it. :)

src/JWT.Extensions.AspNetCore/JwtAuthenticationOptions.cs
drusellers: So the base class has the same property but it returns type `object` as seen here: https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Core/src/AuthenticationSchemeOptions.cs#L33

You can see that Microsoft does the same thing with their OAuth https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/OAuth/src/OAuthOptions.cs#L70

So, I'm mostly copying this pattern. This allows you to add custom events that work just for you.

src/JWT.Extensions.AspNetCore/JwtAuthenticationOptions.cs
drusellers: Because we aren't and can't use auto-properties - we have do initialize in the ctor.

drusellers:
> Why `options.EventsType` is of type `Type`? How it's used later?

It's of type `Type` because its used by the Microsoft DI to resolve an instance of the Type from the container. This is the crux of the work. Allowing me to set the instance of `Events` to an instance out of DI that is of type `EventsType` lets me customize it and get access to things like DbContext.

drusellers:
```csharp
.AddJwt(JwtAuthenticationDefaults.AuthenticationScheme, options =>
{
    options.OnSuccessfulTicket = (logger, ticket) =>
    {
        // ARGH - I can get at the DbContext
    };
});
// later
builder.Services.AddTransient<DoTheWorkHereAsAWorkAround>();
```

src/JWT.Extensions.AspNetCore/JwtAuthenticationOptions.cs
abatishchev: Fair! I'm fine with doing this as long as this is a common pattern

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,101 @@
+namespace JWT.Extensions.AspNetCore
+{
+
```
abatishchev, line 3: nit: extra line break (here and in few other places too)

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+    using System;
+    using Microsoft.AspNetCore.Authentication;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Net.Http.Headers;
+
+    public class JwtAuthenticationEvents
```
abatishchev, line 9: `sealed`, please. Tiny but a perf gain

abatishchev (changes requested):

abatishchev:
Sounds like the right strategy then. Thank you again! Please add some tests and other paperwork (version bump, changelog)

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
drusellers: So, we don't want this sealed because then we can inherit from it. That is the crux of the work. You inherit from the existing one. Then you override the method you are interested in to plug in your own logic.

abatishchev: Ah, right! Makes sense.

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,100 @@
+namespace JWT.Extensions.AspNetCore
+{
+    using System;
```
abatishchev, line 3: Can you move the uisnfa outside the namespace?

drusellers: I'll do a whole formatting pass at the end. :) my IDE is just putting things in my normal spots. But before we sign this puppy off we can square all of the formatting away. No problem. :)

drusellers:
@abatishchev ok, so the next step that I'd like to discuss is moving the events to a "context" approach - but I'm also happy to make that a different PR. The idea is to copy what we see in other authentication handlers like you see here: 
https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/OAuth/src/OAuthHandler.cs#L256

The big things that I think this adds which is nice. 

1. moving to a single context parameter allows us to add more data to the context without breaking the method signature
2. we can provide the HttpContext which has IoC on it via `RequestServices` so people can use the lighter `OnXxxx` methods before a full upgrade to an IoC type.
3. we can also pass the `AuthenticationProperties` which is a common data bag for the authentication pipeline.

I hesitate to take this move in this PR because it would break the API. I can get what I need with just what we have. But to bring it even closer to the GoogleAuthenticationHandler this would be the next step I'd like to do - but happy to do it in another PR.

Also, when we are done, I can squash all of these commits into one clean commit. Is that preferred?

abatishchev:
I like the approach (especially given it's a pattern) and don't mind a breaking change (ideally of course we just mark old things as `[Obsolete]` first. We would just need to jump the major version

drusellers:
Ok, it'll be a bit before I get to it but I'll put the context objects in. I'll chew on how to best make the old stuff still work. It should totally be doable. Just need to think on it. :)

abatishchev:
We also can check in the current PR, just without releasing the package  Either way, sounds good to me!

abatishchev:
Or what I usually did earlier, bump the major version but suffic it with let's say `beta1`. And 2, and so on until it's stable enough.

drusellers:
@abatishchev ok, with this latest push we have context classes with more data available. YAY! The old methods are still supported, although right now its either/or. You can't do both. I'd be game to do a `beta` especially if that turns into a NuGet I can pull in. Then I can run this in my system for a bit to look for any other goodies.

src/JWT.Extensions.AspNetCore/FailedTicketContext.cs
```diff
@@ -0,0 +1,21 @@
+namespace JWT.Extensions.AspNetCore;
```
abatishchev, line 1: Exploit namespace, please

src/JWT.Extensions.AspNetCore/FailedTicketContext.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+
+public class FailedTicketContext
+{
+    public FailedTicketContext(ILogger logger, Exception exception, HttpContext context, JwtAuthenticationOptions options)
```
abatishchev, line 9: Why do we need a separate class for the failed scenario?

src/JWT.Extensions.AspNetCore/FailedTicketContext.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        this.Logger = logger;
+        this.Exception = exception;
+        this.Context = context;
+        this.Options = options;
+    }
+
+    public ILogger Logger { get; }
```
abatishchev, line 17: Line breaks after each property, please

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,103 @@
...
+            };
+
+        public Func<ILogger, string, AuthenticateResult> OnEmptyHeader { get; set; } = (logger, header) =>
+        {
+            _logEmptyHeader(logger, null);
+            return AuthenticateResult.NoResult();
+        };
+
+        public Func<SuccessfulTicketContext, AuthenticateResult> OnSuccessfulTicket { get; set; } =
+            (context) =>
```
abatishchev, line 65: No need for `()`

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -0,0 +1,103 @@
...
+            return AuthenticateResult.Fail(context.Exception);
+        };
+
+        public virtual AuthenticateResult SuccessfulTicket(SuccessfulTicketContext context)
+        {
+            return OnSuccessfulTicket(context);
+        }
+
+        public virtual AuthenticateResult FailedTicket(FailedTicketContext context)
+        {
```
abatishchev, line 83: The `=>` syntax for brevity

abatishchev (changes requested):

src/JWT.Extensions.AspNetCore/FailedTicketContext.cs
drusellers: For this pattern, I'm copying what Microsoft has done with their events classes.

- https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Cookies/src/CookieAuthenticationEvents.cs#L115
- https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/OAuth/src/Events/OAuthEvents.cs#L30
- https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.Okta/OktaAuthenticationHandler.cs#L57

abatishchev: Sorry, I don't get it. I mean I get that you're following the pattern and it's the right thing to do. But the links don't include `FailedTicketContext`

drusellers: Oh, for sure! The biggest reason is that the FailedTicketContext has an Exception property, an the SuccessfulTicketContext has an AuthenticationTicket property. If I had to use one class to represent both, I would then need to make `Ticket` and `Exception` properties nullable. Dedicated classes allow each property to be non-null, which makes sense because in the happy path we ALWAYS have a Ticket and NEVER an error. And vice versa in the other path.

That is the biggest reason from a DX / API design stand point.

Really, the secondary reason is that its a pattern we see in the others.

The links don't include a `FailedTicketContext` since that is _OUR_ class. But for example:

[Cookie](https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Cookies/src/CookieAuthenticationEvents.cs#L115)
- CookieValidatePrincipalContext
- CookieSlidingExpirationContext
- CookieSigningInContext
- CookieSignedInContext
- CookieSigningOutContext
- RedirectContext<CookieAuthenticationOptions>

[OAuth](https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/OAuth/src/Events/OAuthEvents.cs#L30)
- OAuthCreatingTicketContext
- RedirectContext<OAuthOptions>
- which inherits from [RemoteAuthenticationEvents](https://github.com/dotnet/aspnetcore/blob/main/src/Security/Authentication/Core/src/Events/RemoteAuthenticationEvents.cs#L9)
    - AccessDeniedContext
    - RemoteFailureContext
   - TicketReceivedContext

Since I see this pattern of very explicit context classes for each event, I chose to follow that path. I should probably go through and convert our other events to use context objects as well. 

I'm also interested in taking a pass through the Authentication Handles so we can also get AuthenticationProperties passed through as well. 

I hope that I addressed your question. :)

abatishchev: Yes, it does. Thank you!

drusellers:
Ok, did a review of `AuthenticationProperties` usage, and since we don't have multiple "events" that need to pass data between them we can avoid that step for now. winning. Ok, going to make sure I have the  formatting changes cleaned up and will do the other punch list items.

drusellers:
@abatishchev ok, updated change log - saw you got the version numbers. Feeling pretty good right now. Any other items you want me to square away?

drusellers:
@abatishchev do we have a way to cut a beta nuget that I can put into my projects and get some "burn in time" with it?

abatishchev:
Yes, sure. Every check (build) procudes a NuGet package you can download and consume (from the disk). Once merged, I'll publish it to NuGet.org too. But would be great if you could give it a local try first!

abatishchev:
@drusellers I rebased your branch on the latest main after I ran into a few merge conflicts. Can you please check if I missed anything?

abatishchev:
All tests passed, I'm ready to check it in

abatishchev (approved):

abatishchev (approved):

drusellers:
@abatishchev sorry for the delay, was heads down on some other projects. I'll look to grab the latest and report back. TY!

PR #497, opened 2024-04-16
Removed referencing System.Text.Json when targeting modern .NET

### LINKED ISSUE #496: System.Text.Json should only be a dependency for .NET standard
**Describe the problem you are trying to solve.**
I do not want STJ listed as a dependency for net 6

**Available workarounds**
Accept un-necessary dependency

**Additional context**
N/a

thompson-tomo:
## Description

Resolves #496.

This removes System.Text.Json as a dependency for dotnet 6 due to it being natively provided by the framework hence the explicit dependency is not need for that TFM.

## Checklist

- <s>[ ] Tests added</s>
- [x] Version bumped
- [x] Changelog updated


abatishchev:
Can you cite any docs which would confirm that STJ is available without referencing it? If it is then which version? How to update it in the future, e.g. in case of a security vulnerability?

thompson-tomo:
@abatishchev easiest way to confirm it is look in Visual Studio, and expand out the framework under dependencies and that will reveal what is included in the framework. In the case of security vulnerability this made even easier to manage as when the framework used to compile the application gets a new release of STJ (within the same major), the client applications automatically get provided this based on their framework. Hence less maintenance effort.

abatishchev:
Let me check with the .NET team :) I'll circle back.

abatishchev:
Spoke to the .NET team, was told this:

> Don't reference it unless for the TFMs you target that don't have it implicitly.

Thanks for educating me, I dind't know. Will re-review the PR and get it going soon.

abatishchev:
@thompson-tomo can you please check why the build fails?

thompson-tomo:
@abatishchev have tweaked the conditions which should have resolved the build issue

abatishchev (dismissed):

abatishchev (approved):

abatishchev:
Awesome, thank you again, @thompson-tomo!

abatishchev:
Published all packages to NuGet.org as `beta1`

PR #501, opened 2024-05-28
Made events async

drusellers:
## Description

I was integrating the beta build into my project and noticed that I failed to make the events async/await able. This is preventing me from digging into my database to verify attributes about the JWT token in my custom Event Type.

I was careful to avoid doing any actual awaits in the framework to try and minimize the amount of "task switching" that the internal framework is doing.  

## Checklist

- [x] Tests added
- [x] Version bumped
- [x] Changelog updated

### DIFF
--- a/src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
+++ b/src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.Extensions.Logging;
 using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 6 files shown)


src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
```diff
@@ -41,47 +42,47 @@ static JwtAuthenticationEvents()
...
-                return AuthenticateResult.NoResult();
+                return Task.FromResult(AuthenticateResult.NoResult());
             };
 
-        public Func<ILogger, string, AuthenticateResult> OnEmptyHeader { get; set; } = (logger, header) =>
+        public Func<ILogger, string, Task<AuthenticateResult>> OnEmptyHeader { get; set; } = (logger, header) =>
         {
             _logEmptyHeader(logger, null);
-            return AuthenticateResult.NoResult();
+            return Task.FromResult(AuthenticateResult.NoResult());
```
abatishchev, line 62: qq: What's the difference between `AuthenticateResult.None` and `AuthenticateResult.NoResult()`? Would the former a bit more efficient?

abatishchev (approved):

src/JWT.Extensions.AspNetCore/JwtAuthenticationEvents.cs
drusellers: [`AuthenticateResult.None`](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Authentication.Abstractions/src/AuthenticateResult.cs#L50) is an instance boolean property, rather than a factory method.

[`AuthenticateResult.NoResult()`](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Authentication.Abstractions/src/AuthenticateResult.cs#L89) is a factory method that returns a [static _noResult](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Authentication.Abstractions/src/AuthenticateResult.cs#L14) which should avoid an allocation.

So, NoResult appears to be the correct choice, and the ASP.Net team has already done us a solid in helping to keep things efficient. :)

abatishchev:
Released to NuGet.org as [11.0.0-beta4](https://www.nuget.org/packages/JWT.Extensions.AspNetCore/11.0.0-beta4)

PR #509, opened 2024-12-03
Updated System.Text.Json to version 9.0.0

runenilsenoe:
Upgraded Nuget ref for System.Text.Json

[JsonExtensionData] is not used, but its annoying that there are vulnerabilities in refs


abatishchev (approved):

abatishchev:
Thanks for the contribution!!

